PR #50, opened 2026-02-24
feat: add real-world benchmark tasks and real adapter support

shaun0927:
## Summary
- New `realworld` benchmark category with 4 task suites: multi-site crawl, heavy JS execution, pipeline (execute_plan), and scalability curve
- Real adapter support (`--mode real`) for both `run.ts` and `run-parallel.ts` benchmark runners — spawns actual MCP server subprocess
- Helper utilities (`extractTabId`, `extractWorkerTabIds`, `parseResultJson`) for parsing real adapter responses
- Updated README with real mode docs and realworld category description

## Files Changed
- `tests/benchmark/tasks/realworld-crawl.ts` (new) — multi-site crawl benchmark
- `tests/benchmark/tasks/realworld-heavy-js.ts` (new) — heavy JS execution benchmark
- `tests/benchmark/tasks/realworld-pipeline.ts` (new) — pipeline/execute_plan benchmark
- `tests/benchmark/tasks/realworld-scalability.ts` (new) — scalability curve benchmark
- `tests/benchmark/run-parallel.ts` — realworld category + real adapter integration
- `tests/benchmark/run.ts` — `--mode real` support
- `tests/benchmark/tasks/index.ts` — export new task creators
- `tests/benchmark/utils.ts` — response parsing helpers
- `README.md` — benchmark documentation updates

## Test plan
- [ ] Build passes (`npm run build`)
- [ ] `npx ts-node tests/benchmark/run-parallel.ts --category realworld --runs 1` runs in stub mode
- [ ] `npx ts-node tests/benchmark/run-parallel.ts --mode real --category realworld --runs 1` runs with real adapter (requires Chrome)
- [ ] `npm run benchmark -- --mode real` runs sequential benchmarks with real adapter

### DIFF
--- a/tests/benchmark/run-parallel.ts
+++ b/tests/benchmark/run-parallel.ts
@@ -69,10 +69,6 @@ async function main(): Promise<void> {
     : new OpenChromeStubAdapter({ mode: 'ax' });
   const runner = new BenchmarkRunner({ runsPerTask: runs });
 
-  if ('setup' in adapter) {
-    await (adapter as OpenChromeRealAdapter).setup();
-  }
-
   const executePlanPair = createExecutePlanBenchmarkPair();
 
   if (category === 'multistep' || category === 'all') {
@@ -167,10 +163,6 @@ async function main(): Promise<void> {
     console.log(toJSON(categories));
   } else {
     console.log(formatParallelReport(categories));
-  }
-
-  if ('teardown' in adapter) {
-    await (adapter as OpenChromeRealAdapter).teardown();
   }
 }
 
--- a/tests/benchmark/run.ts
+++ b/tests/benchmark/run.ts
@@ -46,11 +46,6 @@ async function main(): Promise<void> {
     ? new OpenChromeRealAdapter({ mode: 'dom' })
     : new OpenChromeAdapter({ mode: 'dom' });
 
-  if (mode === 'real') {
-    await (axAdapter as OpenChromeRealAdapter).setup();
-    await (domAdapter as OpenChromeRealAdapter).setup();
-  }
-
   console.log(`Running benchmarks in AX mode (${mode})...`);
   const axReport = await runner.run(axAdapter);
 
@@ -79,10 +74,6 @@ async function main(): Promise<void> {
     console.log(BenchmarkRunner.formatReport(reports));
   }
 
-  if (mode === 'real') {
-    await (axAdapter as OpenChromeRealAdapter).teardown();
-    await (domAdapter as OpenChromeRealAdapter).teardown();
-  }
 }
 
 main().catch((err) => {


shaun0927 (commented):
## PR #50: feat: add real-world benchmark tasks and real adapter support

None.

- [ ] **[P1]** Double `setup()`/`teardown()` calls on `OpenChromeRealAdapter` in both `run.ts` and `run-parallel.ts` — `tests/benchmark/run.ts:49-52` and `tests/benchmark/run-parallel.ts:72-74` (Confidence: 95/100)
  - Impact: `BenchmarkRunner.run()` unconditionally calls `adapter.setup()` and `adapter.teardown()` if those methods exist (see `benchmark-runner.ts` lines 121 and 132). `run.ts` already calls `setup()` explicitly before `runner.run()` and `teardown()` after. This causes the MCP server subprocess to be spawned twice per adapter — the second `setup()` call opens a second child process while the first is still live. The second `teardown()` then calls `oc_stop` and `kill` on an already-terminated process, causing unhandled errors or silent failures.
  - Fix: Either (a) remove the explicit `setup()`/`teardown()` calls from `run.ts` and `run-parallel.ts` and rely on `BenchmarkRunner.run()` to manage them, or (b) remove the `setup`/`teardown` methods from the `MCPAdapter` interface so the runner does not call them, and keep the explicit calls in the script entry points. Option (a) is simpler. Same issue exists in `run-parallel.ts` at lines 72-74 and 194-196.

- [ ] **[P2]** Inline counter literals miss `serverTimingMs` in sequential tasks — `tests/benchmark/tasks/realworld-crawl.ts:336`, `tests/benchmark/tasks/realworld-heavy-js.ts:520`, `tests/benchmark/tasks/realworld-pipeline.ts:677`, `tests/benchmark/tasks/realworld-scalability.ts:813`
  - Suggestion: Use `createCounters()` (already exported from `utils.ts`) instead of the inline `{ inputChars: 0, outputChars: 0, toolCallCount: 0 }` literals. `measureCall` detects the absence of `serverTimingMs` and silently skips accumulation, meaning server-side execution timing is not captured for sequential tasks in real mode. The parallel tasks in the same files correctly use `createCounters()`.

- [ ] **[P2]** `parseResultJson` is not exported from `utils.ts` despite being listed as a key helper in the PR summary — `tests/benchmark/utils.ts:56`
  - Suggestion: Either export it for use by future task files, or update the PR description to clarify it is intentionally private. Current state is consistent internally but the PR summary creates a misleading impression.

- [ ] **[P2]** Hardcoded `https://httpbin.org/delay/1` and `https://httpbin.org/delay/2` in scalability benchmarks mean the realworld suite cannot run in restricted network environments (CI, offline, air-gapped) — `tests/benchmark/tasks/realworld-scalability.ts:797`
  - Suggestion: Add a guard or README note that `--category realworld` requires network access. The parallel runner README entry already states this, but a runtime check or `--skip-network` flag would improve DX.

- [ ] **[P2]** `extractWorkerTabIds` type annotation on inner `.every()` callback is misleading — `tests/benchmark/utils.ts:93`
  - Suggestion: Change `ids.every((id: string) => typeof id === 'string')` to `ids.every((id: unknown) => typeof id === 'string')` for correctness. The current annotation gives false type safety since the array element type is already widened to `string` via the `.map()` cast, but `id` could be `undefined` if `w.tabId` was absent.

### Summary

| Priority | Count |
|----------|-------|
| P0 | 0 |
| P1 | 1 |
| P2 | 4 |

### Verdict

⚠️ REQUEST_CHANGES — One P1 issue: the double `setup()`/`teardown()` lifecycle bug will cause duplicate MCP subprocess spawning in real mode. The fix is straightforward (remove explicit lifecycle calls from entry points and let `BenchmarkRunner.run()` manage them, or vice versa). All P2 items are minor and can be addressed as follow-ups.

### Merge Notes

- No file conflicts with other open PRs detected.
- This PR targets `develop` (correct per project convention).
- The `realworld` category is correctly excluded from `--category all`, preventing unintended network calls in CI.

PR #63, opened 2026-02-25
fix: stop killing user's Chrome — use temp profile by default

### LINKED ISSUE #62: fix: stop killing user's Chrome — use temp profile by default when Chrome is running
## Problem

When Chrome is running without a debug port (the normal way users open Chrome), OpenChrome **quits Chrome** to get exclusive access to the real profile directory. This is destructive — the user loses all their open tabs and windows.

### Current behavior

```
ensureChrome() decision tree:

1. Chrome running WITH debug port → connect (ideal, no disruption)
2. Chrome running WITHOUT debug port → QUIT CHROME → relaunch with debug port
3. Chrome not running → launch with debug port
```

Case 2 is the problem. The graceful restart at `launcher.ts:217-229` calls `quitAndUnlockProfile()` which sends AppleScript quit / SIGTERM to Chrome, waits for it to exit, then relaunches with the real profile.

### Why this happens

Chrome enforces exclusive profile ownership via lock files (`SingletonLock`, `SingletonSocket`, `SingletonCookie`). Two Chrome instances cannot share the same `--user-data-dir`. The current logic prioritizes reusing the real profile (full cookie/extension access) over preserving the user's running Chrome.

### Existing solution buried as fallback

A temp profile + cookie copy path already exists at `launcher.ts:280-286` but only activates when graceful restart **fails** (e.g., Chrome refuses to quit within 10s). This path:
1. Creates a temp directory
2. Copies `Cookies`, `Local State`, and `Preferences` from the real profile
3. Launches a new Chrome instance on the temp profile with `--remote-debugging-port`

This works well — cookies are decryptable because the same OS user's keychain is used.

## Solution

Make temp profile the **default** when Chrome is already running, instead of killing it.

### Changes

1. **`src/chrome/launcher.ts`**: Remove `quitAndUnlockProfile()` from the default `ensureChrome()` path. When profile is locked and Chrome is running, always fall through to the temp profile + cookie copy path.

2. **New `--restart-chrome` CLI flag**: For users who explicitly want the old behavior (kill Chrome and use real profile), add an opt-in flag.

3. **Log message**: When using temp profile, log a clear message explaining cookies are a snapshot.

### New decision tree

```
ensureChrome():
  Chrome on debug port? → connect (unchanged)
  autoLaunch=false?     → throw (unchanged)
  Profile locked + Chrome running?
    --restart-chrome flag? → quit Chrome, use real profile (old behavior, opt-in)
    default?              → temp profile + cookie copy (NEW default)
  Profile not locked?     → use real profile directly (unchanged)
```

### Trade-offs

| Aspect | Old (restart) | New (temp profile) |
|--------|---------------|-------------------|
| User's Chrome | Killed | **Preserved** |
| Cookies | Full real profile | Snapshot at launch time |
| Extensions | Available | Not available (--disable-extensions) |
| New logins in Chrome | Reflected | Not reflected until restart |
| New logins in OpenChrome | Persisted | Lost on exit |

The cookie snapshot trade-off is acceptable because:
- Most users have already logged into their important sites
- The alternative (Chrome being killed) is far worse UX
- Users who need real profile access can use `--restart-chrome`

## Files to modify

- `src/chrome/launcher.ts` — Change `ensureChrome()` decision logic
- `src/config/global.ts` — Add `restartChrome` config option
- `src/index.ts` — Add `--restart-chrome` CLI flag
- `tests/chrome/launcher.test.ts` — Update tests for new default behavior

shaun0927:
## Summary

Closes #62

- **Default behavior changed**: When Chrome is running without a debug port, OpenChrome now uses a temp profile with cookie copy instead of killing the user's Chrome
- **Opt-in restart**: Added `--restart-chrome` CLI flag for users who want the old behavior (quit Chrome to reuse real profile)
- **No breaking changes**: Users who don't pass `--restart-chrome` get the safe default automatically

## Changes

| File | Change |
|------|--------|
| `src/chrome/launcher.ts` | Guard restart branch behind `restartChrome` opt-in flag |
| `src/config/global.ts` | Add `restartChrome` to `GlobalConfig` interface |
| `src/index.ts` | Add `--restart-chrome` CLI option to `serve` command |
| `tests/chrome/launcher-restart.test.ts` | 4 new tests for default/opt-in behavior + robust mocking |

## Before / After

| Scenario | Before | After |
|----------|--------|-------|
| Chrome running, no debug port | Kills Chrome, reuses real profile | Uses temp profile + cookie copy |
| Chrome running + `--restart-chrome` | N/A | Kills Chrome, reuses real profile (old behavior) |
| Chrome not running | Uses real profile | Uses real profile (unchanged) |

## Test plan

- [x] Build passes (0 errors)
- [x] All 1066 tests pass (51 suites)
- [x] 13 launcher-restart tests pass (including 4 new)
- [ ] Manual: `openchrome serve --auto-launch` with Chrome open — Chrome should NOT be killed
- [ ] Manual: `openchrome serve --auto-launch --restart-chrome` with Chrome open — Chrome restarts

### DIFF
--- a/src/chrome/launcher.ts
+++ b/src/chrome/launcher.ts
@@ -25,6 +25,8 @@ export interface LaunchOptions {
   autoLaunch?: boolean;
   /** If true, force using a temp directory instead of real Chrome profile */
   useTempProfile?: boolean;
+  /** If true, quit running Chrome to reuse the real profile (default: false — uses temp profile instead) */
+  restartChrome?: boolean;
 }
 
 const DEFAULT_PORT = 9222;
@@ -211,21 +213,19 @@ export class ChromeLauncher {
       );
     }
 
-    // Graceful restart: if Chrome is running without debug port and profile is locked,
-    // quit Chrome gracefully so it saves session state, then let the existing launch
-    // logic detect the unlocked real profile and relaunch with --remote-debugging-port.
-    if (!options.useTempProfile) {
+    // Graceful restart: only when explicitly opted in via --restart-chrome flag.
+    // Default behavior: skip restart, fall through to temp profile + cookie copy.
+    const restartChrome = options.restartChrome ?? getGlobalConfig().restartChrome ?? false;
+    if (!options.useTempProfile && restartChrome) {
       const realProfileDir = this.getRealChromeProfileDir();
       if (realProfileDir && this.isProfileLocked(realProfileDir) && this.isChromeRunning()) {
-        console.error('[ChromeLauncher] Chrome running without debug port — attempting graceful restart...');
+        console.error('[ChromeLauncher] --restart-chrome: attempting graceful restart...');
         const unlocked = await this.quitAndUnlockProfile(realProfileDir);
         if (unlocked) {
           console.error('[ChromeLauncher] Chrome quit successfully, profile unlocked. Relaunching with debug port...');
         } else {
           console.error('[ChromeLauncher] Graceful restart failed, falling back to temp profile...');
         }
-        // On success: existing launch logic below detects unlocked profile
-        // On failure: falls through to existing temp-profile path
       }
     }
 
--- a/src/config/global.ts
+++ b/src/config/global.ts
@@ -15,6 +15,8 @@ export interface GlobalConfig {
   useHeadlessShell?: boolean;
   /** Run Chrome in headless mode (default: true when auto-launch is enabled) */
   headless?: boolean;
+  /** If true, quit running Chrome to reuse the real profile instead of using temp profile (default: false) */
+  restartChrome?: boolean;
   /** Chrome Pool settings for managing multiple Chrome instances */
   pool?: {
     /** Enable the Chrome pool (default: true) */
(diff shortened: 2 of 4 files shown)


shaun0927 (commented):
## Code Review

**Verdict: APPROVE** ✅ (P0: 0, P1: 1, P2: 1)

### P1 — Silent flag conflict between `restartChrome` and `useTempProfile`

When both flags are true, `useTempProfile` silently wins. Consider adding a log when both conflict:
```typescript
if (restartChrome && options.useTempProfile) {
  console.error('[ChromeLauncher] --restart-chrome ignored: useTempProfile is set');
}
```

### P2 — Missing test for config-level `restartChrome` fallback

Test mock for `getGlobalConfig` returns `{}`, so the middle layer of the `??` chain is never tested.

### Positives
- Safe default: Chrome is no longer killed without explicit opt-in
- Clean three-level fallback chain with `??` (not `||`)
- Build passes with all 1063 tests green

---

PR #79, opened 2026-02-25
fix: support top-level await in javascript_tool

### LINKED ISSUE #69: fix: javascript_tool fails on top-level await — eval() not wrapped in async IIFE
## Problem

LLMs frequently generate JavaScript with `await` (e.g., `await new Promise(r => setTimeout(r, 500))`) in `javascript_tool`, which fails with:

```
Error: JavaScript error: await is only valid in async functions and the top level bodies of modules
```

This is a recurring pattern that causes workflow interruption and LLM confusion.

## Root Cause

`src/tools/javascript.ts` line 66 uses indirect eval:

```typescript
let evalResult = (0, eval)(jsCode);
```

While the outer wrapper IS async (line 63), `eval()` executes the user's code string at the top-level scope where `await` is not valid. The tool already handles Promise results (lines 69-70) but cannot handle `await` syntax in the input code.

**What works:**
```javascript
fetch(url)                    // ✅ Returns Promise, handled by line 69-70
Promise.resolve(123)          // ✅ Returns Promise
```

**What fails:**
```javascript
await new Promise(r => setTimeout(r, 500))   // ❌ await at eval top-level
const data = await fetch(url)                 // ❌ await at eval top-level
```

## Proposed Solution

Wrap user code in an async IIFE:

```typescript
// Change line 66 from:
let evalResult = (0, eval)(jsCode);

// To:
let evalResult = (0, eval)(`(async () => { ${jsCode} })()`);
```

This makes `await` valid inside the async function body while preserving backward compatibility:
- Existing non-await code still works (executed inside async function, last expression returned)
- Promise-returning code still works (outer handler awaits the IIFE's Promise)
- `await` syntax now works natively

### Edge case: expression-only code

For code that's just an expression (e.g., `document.title`), the async IIFE wrapping needs to handle both statement and expression modes:

```typescript
// Detect if code is a single expression or statements
const isExpression = !jsCode.includes(';') && !jsCode.includes('\n');
let evalResult;
if (isExpression) {
  evalResult = (0, eval)(jsCode);  // simple expression, no wrapping needed
} else {
  evalResult = (0, eval)(`(async () => { ${jsCode} })()`);
}
```

Or use a `return` heuristic — if the last statement is an expression, add implicit `return`.

## Acceptance Criteria

- [ ] `await new Promise(r => setTimeout(r, 500))` works in javascript_tool
- [ ] `const resp = await fetch(url); resp.status` works
- [ ] Existing non-await code (`document.title`, `1+1`) still works
- [ ] Error messages remain clear for actual JS syntax errors

shaun0927:
## Summary

Closes #69

- **Problem**: LLMs frequently generate JavaScript with `await` (e.g., `await new Promise(...)`, `const data = await fetch(url)`) which fails in `javascript_tool` because indirect eval `(0, eval)(code)` does not support `await` at the top-level scope.
- **Solution**: Implements a 3-tier eval strategy with automatic fallback:
  1. **Tier 1** — Direct eval (backward-compatible, handles all non-await code)
  2. **Tier 2** — Async IIFE expression wrapper for single await expressions (e.g., `await fetch(url)`)
  3. **Tier 3** — Async IIFE statement wrapper with return-last-expression heuristic for multi-statement await code (e.g., `const resp = await fetch(url); resp.status`)
- Updates tool description to advertise top-level await support

## Acceptance Criteria

- [x] `await new Promise(r => setTimeout(r, 500))` works in javascript_tool
- [x] `const resp = await fetch(url); resp.status` works
- [x] Existing non-await code (`document.title`, `1+1`) still works
- [x] Error messages remain clear for actual JS syntax errors

## Test Plan

- [x] 8 new test cases added in `Top-level Await Support` describe block
- [x] All 29 tests pass (21 existing + 8 new)
- [x] Type check passes (zero new errors)
- [x] Tested patterns: single await, setTimeout, multi-statement, non-await code, string expressions, syntax errors, null resolution, multiple sequential awaits

### DIFF
--- a/src/tools/javascript.ts
+++ b/src/tools/javascript.ts
@@ -79,10 +79,14 @@ const handler: ToolHandler = async (
           } catch {
             // Tier 3: Multi-statement with return-last-expression heuristic
             const lastSemiIdx = trimmed.lastIndexOf(';');
-            if (lastSemiIdx !== -1) {
+            // Guard: skip split if the semicolon is likely inside a string literal
+            // (odd number of unescaped quotes after the split point)
+            const afterSplit = lastSemiIdx !== -1 ? trimmed.substring(lastSemiIdx + 1) : '';
+            const quotesSafe = ((afterSplit.match(/(?<!\\)["'`]/g) || []).length % 2) === 0;
+            if (lastSemiIdx !== -1 && quotesSafe) {
               const head = trimmed.substring(0, lastSemiIdx + 1);
-              const tail = trimmed.substring(lastSemiIdx + 1).trim();
-              if (tail && !/^(const|let|var|function|class|if|for|while|switch|try|throw|return)\b/.test(tail)) {
+              const tail = afterSplit.trim();
+              if (tail && !/^(const|let|var|function|class|if|for|while|do|switch|try|throw|return|delete|typeof|void)\b/.test(tail)) {
                 evalResult = (0, eval)(`(async()=>{${head}\nreturn(${tail});})()`);
               } else {
                 evalResult = (0, eval)(`(async()=>{${trimmed}})()`);


shaun0927 (commented):
## Code Review

**Verdict: REQUEST_CHANGES** (P0: 0, P1: ~2, P2: ~2)

### P1-1: Tier 3 "return last expression" heuristic — missing keywords in blocklist
The regex `/^(const|let|var|function|class|if|for|while|switch|try|throw|return)\b/` for detecting statements is missing `do`, `import`, `export`, `delete`, `typeof`, `void`. These would cause the heuristic to incorrectly wrap them with `return`.

### P1-2: Semicolon-based splitting is fragile with strings/template literals
`trimmed.lastIndexOf(';')` will incorrectly split on semicolons inside string literals (e.g., `await fetch("url;param=1"); x.json()`).

### Positives
- Three-tier detection approach (explicit return, IIFE, heuristic) is well-designed
- Good test coverage for common patterns
- No `console.log` in tool handlers
- TypeScript compiles cleanly

---

PR #83, opened 2026-02-25
fix: validate ref identity before click to prevent stale ref misclicks

### LINKED ISSUE #70: fix: find tool returns stale/wrong refs — DOM mutation between search and backendNodeId capture
## Problem

The `find` tool sometimes returns refs that point to **wrong elements**. Observed in session:

- `find` returns `[ref_41] option: "neck-and-neck_19yr98%" at (521, 143)`
- Clicking `ref_41` navigates to `/projects` (a sidebar link), not the dropdown option

This causes the LLM to click wrong elements, disrupting workflow and triggering wandering patterns.

## Root Cause

**DOM mutation window** between element discovery and backendNodeId capture in `src/tools/find.ts` (lines 292-340):

```
Step 1: page.evaluate() → finds & sorts elements, marks with __findIndex
Step 2: Runtime.evaluate → collects marked elements by __findIndex
Step 3: Runtime.getProperties → gets objectIds for each element
Step 4: DOM.describeNode (parallel) → gets backendNodeIds
```

Between Steps 1-4, the DOM can change (dropdown opens/closes, React re-render, CSS transition). When the DOM mutates:
- `objectId` references from Step 3 may point to different elements
- `backendNodeId` from Step 4 is for the **current** element at that reference, not the original

**No validation**: When `computer.ts` clicks a ref, it trusts the stored backendNodeId without verifying it still points to the expected element:

```typescript
// computer.ts — no identity check before click
const backendNodeId = refIdManager.resolveToBackendNodeId(sessionId, tabId, ref);
const { model } = await cdpClient.send(page, 'DOM.getBoxModel', { backendNodeId });
// Clicks whatever element is at that backendNodeId — could be wrong!
```

**Ref invalidation gaps**:
- Refs ARE cleared on page navigation and `read_page` calls
- Refs are NOT cleared on DOM mutations (dropdown open/close, re-render)
- No staleness check before using a ref

### 1. Atomic ref capture (reduce mutation window)

Combine element discovery + backendNodeId capture into a single CDP call:

```typescript
// Instead of 4 separate steps, use a single evaluate that returns
// both element info AND a stable reference
const results = await page.evaluate(() => {
  // Find, score, and collect all info in one synchronous pass
  return elements.map(el => ({
    text: el.textContent,
    tagName: el.tagName,
    rect: el.getBoundingClientRect(),
    // Store element reference for immediate backendNodeId resolution
  }));
});
```

### 2. Ref identity validation before click

```typescript
// In computer.ts, before clicking a ref:
const stored = refIdManager.getRefEntry(sessionId, tabId, ref);
const { node } = await cdpClient.send(page, 'DOM.describeNode', { backendNodeId: stored.backendDOMNodeId });

// Verify element identity
if (node.nodeName.toLowerCase() !== stored.tagName ||
    !node.attributes?.includes(stored.identifyingAttr)) {
  return { isError: true, content: [{
    type: 'text',
    text: `Ref ${ref} is stale — element has changed. Run find again.`
  }]};
}
```

### 3. Store element fingerprint with ref

```typescript
// In ref-id-manager.ts, extend RefEntry:
interface RefEntry {
  refId: string;
  backendDOMNodeId: number;
  role: string;
  name: string;
  tagName: string;       // NEW: for validation
  textContent: string;   // NEW: for validation
  createdAt: number;
}
```

### 4. TTL-based ref expiration

```typescript
// Refs older than 30s are considered potentially stale
const REF_TTL_MS = 30000;
if (Date.now() - stored.createdAt > REF_TTL_MS) {
  console.warn(`Ref ${ref} is ${age}ms old — may be stale`);
}
```

## Acceptance Criteria

- [ ] Clicking a ref always clicks the element that was displayed in `find` results
- [ ] Stale refs produce a clear error message instead of clicking wrong elements
- [ ] Ref identity validated before click (tagName + text match)
- [ ] DOM mutations between `find` and `click` detected and reported

shaun0927:
## Summary

Fixes #70 — `find` tool returns stale/wrong refs when DOM mutates between element discovery and `backendNodeId` capture, causing clicks on wrong elements.

### Root Cause
Between `find` discovering an element and `computer` clicking it, the DOM can change (dropdown open/close, React re-render, CSS transition). The stored `backendNodeId` may point to a different element than what was displayed in `find` results. No validation existed to detect this mismatch.

### Solution

**1. Element fingerprint storage** (`ref-id-manager.ts`)
- Extended `RefEntry` with `tagName` and `textContent` fields
- These are captured at ref creation time and stored alongside the `backendNodeId`

**2. Ref identity validation** (`ref-id-manager.ts`)
- Added `validateRef()` method that compares stored fingerprint against current DOM node
- Added `isRefStale()` method with 30-second TTL-based staleness detection
- Validation checks: tagName match (case-insensitive), textContent prefix match (first 30 chars)

**3. Pre-click validation** (`computer.ts`, `form-input.ts`)
- Before clicking a ref, uses CDP `DOM.describeNode` to get current node properties
- Compares against stored fingerprint via `validateRef()`
- Returns clear error message if stale: `"Error: ref_41 is stale — Element tag changed: expected <option>, found <a>. Run find or read_page again to get fresh refs."`
- Graceful fallback: if validation CDP call fails, proceeds with click (getBoxModel will catch removed elements)

**4. Fingerprint propagation** (`find.ts`, `click-element.ts`, `wait-and-click.ts`)
- All tools that create refs now pass `tagName` and `textContent` to `generateRef()`

### Files Changed

| File | Change |
|------|--------|
| `src/utils/ref-id-manager.ts` | Core: `RefEntry` fingerprint fields, `validateRef()`, `isRefStale()` |
| `src/tools/computer.ts` | Pre-click validation in `resolveRefToCoordinates()` |
| `src/tools/form-input.ts` | Pre-action validation before `DOM.resolveNode` |
| `src/tools/find.ts` | Pass fingerprint to `generateRef()` |
| `src/tools/click-element.ts` | Pass fingerprint to `generateRef()` |
| `src/tools/wait-and-click.ts` | Pass fingerprint to `generateRef()`, extend `FoundElement` |
| `tests/utils/mock-session.ts` | Updated mock with new methods |

## Test plan

- [x] TypeScript compiles with zero errors
- [x] All 145 existing tests pass (ref-id-manager, resolver, computer, form-input, stable-ref)
- [x] Backward compatible — refs without fingerprint (e.g., from `read_page` AX tree) skip validation gracefully
- [ ] Manual: Use `find` to locate a dropdown option, trigger DOM mutation, then click the ref — should get stale ref error instead of wrong click

### DIFF
--- a/src/tools/computer.ts
+++ b/src/tools/computer.ts
@@ -528,6 +528,7 @@ const handler: ToolHandler = async (
       }
 
       case 'scroll_to': {
+        // Note: scroll_to does not validate ref identity (less destructive than click)
         if (!ref) {
           return {
             content: [{ type: 'text', text: 'Error: ref is required for scroll_to' }],


shaun0927 (commented):
## Code Review

**Verdict: REQUEST_CHANGES** (P0: 0, P1: 3, P2: 4)

### P1-1: `read-page.ts` not updated — refs from `read_page` won't have fingerprints
`read_page` is a primary source of refs via the AX tree. Its `generateRef` call was not updated to pass `tagName`/`textContent`, so all `read_page` refs skip validation entirely. Either add fingerprinting or document the exclusion.

### P1-2: `scroll_to` action bypasses validation
`scroll_to` resolves ref directly without going through validation. While less destructive than clicking, it creates an inconsistent safety boundary.

### P1-3: `textContent` validation is dead code
Both callers of `validateRef` never pass `currentTextContent`, making the text comparison unreachable. Wire it up or remove it.

### P2 highlights
- Duplicated validation logic across `computer.ts` and `form-input.ts`
- No unit tests for `validateRef` and `isRefStale`
- `isStale` TTL variable confusing in mismatch case

### Positives
- Core validation logic (fingerprint-at-creation, validate-before-click) is well-designed
- Approach is sound

---

PR #145, opened 2026-03-02
fix: resolve MCP dual-registration and read-page DOM fallback regression

shaun0927:
## Summary

- **Fix MCP setup dual-registration**: `oc setup` now removes openchrome from both `user` and `project` scopes before re-adding, preventing duplicate MCP server entries that cause connection failures
- **Fix read-page DOM fallback**: When DOM serialization fails, the handler gracefully falls through to AX mode instead of returning a raw error string
- **Fix stale test assertion**: Update `update-check.test.ts` to match the current update notification message (`npx openchrome-mcp@latest setup`)

## Context

After the v1.6.1 release, all 9 CI jobs failed due to a single test regression in `read-page.test.ts`. Additionally, users running `oc setup` with different `--scope` values could end up with openchrome registered in both user-level and project-level configs, causing Claude Code to attempt launching two MCP server instances with a Chrome port 9222 conflict.

### Root Causes

**1. MCP dual-registration (`cli/index.ts:111`)**

The `claude mcp remove openchrome` call had no `--scope` flag, so it only removed from whichever scope the CLI defaults to. If a user previously ran `oc setup` at user scope and then `oc setup --scope project`, the first entry survived, creating a duplicate.

**2. read-page DOM fallback (`src/tools/read-page.ts:287-306`)**

The DOM branch called `serializeDOM()` without a try/catch. When serialization failed, the error propagated to the outer catch, returning `"Read page error: DOM serialization failed"` instead of falling back to AX mode with a truncated output and guidance message.

**3. update-check test (`tests/cli/update-check.test.ts:95`)**

The update notification message was changed from `npm update -g` to `npx openchrome-mcp@latest setup` but the test assertion was not updated.

## Changes

| File | Change |
|------|--------|
| `cli/index.ts` | Remove from both `user` and `project` scopes before adding |
| `src/tools/read-page.ts` | Wrap DOM branch in try/catch, fall through to AX on failure |
| `tests/cli/update-check.test.ts` | Update assertion to match current message |

## Test plan

- [x] `read-page.test.ts` — 27/27 passing (was 26/27)
- [x] `update-check.test.ts` — 6/6 passing (was 5/6)
- [x] Build succeeds (`npm run build`)
- [x] No new test regressions introduced
- [ ] CI: verify all 9 matrix jobs pass (macOS/Ubuntu/Windows × Node 18/20/22)

### DIFF
--- a/.claude/commands/release-oc.md
+++ b/.claude/commands/release-oc.md
@@ -19,9 +19,11 @@ git stash list
 git branch -a
 gh pr list --state open --json number,title,headRefName,baseRefName,additions,deletions,author,files
 npm run build
+npm run lint
+npm test
 ```
 
-**Gate**: If build fails, fix build errors first. Do NOT proceed with failing build.
+**Gate**: If build, lint, or tests fail, fix errors first. Do NOT proceed with any failures.
 
 ## STEP 2: Classify Open PRs
 
@@ -122,6 +124,8 @@ Before merging ANY PR, verify ALL of these:
 ```bash
 npm ci                                                 # must pass (lockfile in sync)
 npm run build                                          # must pass
+npm run lint                                           # must pass (no errors)
+npm test                                               # must pass (ALL test suites green)
 git diff --name-only HEAD | wc -l                      # must be 0 (clean tree)
 ```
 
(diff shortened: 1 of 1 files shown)


shaun0927 (commented):
**Self-review: LGTM**

4 files changed, all verified:

1. **`src/tools/read-page.ts`** — DOM serialization wrapped in try/catch, falls through to AX mode on failure. Fixes CI test `falls back to truncated AX output when DOM serialization fails`.

2. **`cli/index.ts`** — `oc setup` now removes from both `user` and `project` scopes before adding, preventing dual-registration MCP conflicts.

3. **`tests/cli/update-check.test.ts`** — Fixed stale assertion (`npm update -g` → `npx openchrome-mcp@latest setup`) to match current output.

4. **`.claude/commands/release-oc.md`** — Added `npm test`, `npm run lint` gates to Steps 1/5/6, added CI verification gate (Step 8a) before publish, updated completion checklist.

P0: 0 | P1: 0 | P2: 0

Issue #246, opened 2026-03-07
BUG:  Ai can't control the browser because of initial startup issue related to profile locking


RezoxP:
`Failure in MCP tool execution: calling "tools/call": Chrome debug port 9222 not available after 60000ms OS: win32 x64 10.0.19045 Chrome: C:\Program Files\Google\Chrome\Application\chrome.exe Profile: C:\Users\User\AppData\Local\Google\Chrome\User Data (real) PID: 5760 Exit code: still running Stderr: fiiekppkjkjjphdpbfkphoea does not exist or is not readable. DevTools remote debugging requires a non-default data directory. Specify this using --user-data-dir. Created TensorFlow Lite XNNPACK delegate for CPU. [5760:7336:0307/221130.023:ERROR:google_apis\gcm\engine\registration_request.cc:290] Registration response error message: DEPRECATED_ENDPOINT [5760:7336:0307/221152.609:ERROR:google_apis\gcm\engine\registration_request.cc:290] Registration response error message: DEPRECATED_ENDPOINT Common causes: - Another Chrome instance is using the same --user-data-dir (profile lock) - Port conflict: another process is bound to port 9222 - Firewall/antivirus blocking localhost connections - Chrome 136+: requires --user-data-dir with --remote-debugging-port`

the browser starts but can't control it. my chrome Version is `145.0.7632.160 (Official Build) (64-bit)`

shaun0927:
## Root Cause Analysis

After thorough investigation, the root cause has been identified. This is a **Chrome 136+ security policy change** combined with **inadequate profile handling** in OpenChrome.

### What Changed in Chrome 136+

Chrome 136 introduced a security restriction: when `--remote-debugging-port` is used, `--user-data-dir` must point to a **non-default** data directory. Passing the user's default Chrome profile path (e.g., `C:\Users\User\AppData\Local\Google\Chrome\User Data`) is now explicitly rejected, even though `--user-data-dir` IS being provided.

### Why This Causes the Error

The failure chain in `launcher.ts`:

1. `isProfileLocked()` on Windows checks for `lockfile` — if Chrome is **not** running, no lockfile → returns `false`
2. `resolveProfile()` sees `isProfileLocked: false` → returns `profileType: 'real'` with the **default** Chrome profile path
3. Chrome is launched with `--user-data-dir=<default-path>` + `--remote-debugging-port=9222`
4. Chrome 136+ detects this is the default directory → **refuses to open the debug port**
[...]

PR #250, opened 2026-03-08
fix: improve Windows profile lock detection with process verification

shaun0927:
## Summary

- Windows `isProfileLocked()` now cross-checks running `chrome.exe` processes when `lockfile` is absent
- Uses a 3-tier fallback strategy: `wmic` (Win10) → PowerShell `Get-CimInstance` (Win11) → `tasklist` (universal)
- For the `tasklist` fallback (can't determine specific profile), assumes locked only when `profileDir` is the default Chrome directory

## Root Cause

The Windows lock detection previously only checked for `lockfile` existence in the user data directory. Chrome on Windows may not always create this file reliably, or it may be in a subdirectory. This caused `isProfileLocked()` to return `false` even when Chrome was running with the same profile, leading to profile conflicts.

## Changes

- `ChromeLauncher.isProfileLocked()`: Added optional `_platform` parameter for testability. After lockfile check, queries process command lines via `wmic` → PowerShell → `tasklist` fallback chain
- All `execSync` calls use `stdio: ['ignore', 'pipe', 'ignore']` and timeouts (5-8s) to prevent hanging
- 7 new tests covering all fallback paths, mocked with real temp directories

## Test plan

- [x] Build passes (`npm run build`)
- [x] All 15 launcher-port-race tests pass (7 new)
- [x] Tests run cross-platform (use `_platform` param instead of mocking `os.platform()`)
- [ ] Manual: verify on Windows 10 (wmic) and Windows 11 (PowerShell fallback)

Ref #246

### DIFF
--- a/src/chrome/launcher.ts
+++ b/src/chrome/launcher.ts
@@ -675,8 +675,8 @@ export class ChromeLauncher {
    * On Unix, validates SingletonLock symlink targets by checking if the PID is alive,
    * so stale lock files from crashed Chrome instances are correctly ignored.
    */
-  private isProfileLocked(profileDir: string, _platform?: string): boolean {
-    const platform = _platform || os.platform();
+  private isProfileLocked(profileDir: string, platformOverride?: string): boolean {
+    const platform = platformOverride || os.platform();
     if (platform === 'win32') {
       // Windows Chrome uses a 'lockfile' in the user data directory
       const lockFile = path.join(profileDir, 'lockfile');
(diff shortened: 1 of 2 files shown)


shaun0927 (commented):
## Code Review — PR #250

**Reviewed by:** oc-code-reviewer (3 agents) + Greptile

### Findings addressed:
- ✅ P1: **Fixed double-escape bug** — wmic path comparison was always failing on real Windows due to `\\` → `\\\\` transformation. Now uses forward-slash normalization on both sides.
- ✅ P2 (Greptile): Renamed `_platform` to `platformOverride` for clarity

### Deferred (accepted risk):
- P1: `execSync` vs `execFileSync` — acceptable for static commands, no user input interpolated
- P1: `quitAndUnlockProfile` polling with long timeouts — separate optimization issue
- P1: `platformOverride` not threaded to `getRealChromeProfileDir` — tests already mock it via spy
- P2: wmic 5s timeout wasted on Windows 11 — future optimization with caching

**Verdict: LGTM** — Critical wmic bug fixed, 15 tests passing.

Issue #257, opened 2026-03-08
Enchanment: Make it more stealthy so it can go through captchas like turnstile and more.


RezoxP:
<img width="351" height="71" alt="Image" src="https://github.com/user-attachments/assets/ef42862a-0268-4a52-adec-8db697d531be" />

Make it more stealthy so it can go through captchas like turnstile and more. currently turnstile can't be solved even with human click in the controled browser

PR #265, opened 2026-03-09
feat: enhance stealth defenses against anti-bot detection (#257)

shaun0927:
## Summary

- Remove five Chrome launch flags identified as bot-detection signals by Patchright's analysis: `--metrics-recording-only`, `--disable-extensions`, `--disable-component-extensions-with-background-pages`, `--disable-default-apps`, and `--disable-ipc-flooding-protection`
- Keep three safe non-signal flags for non-real profiles: `--disable-background-networking`, `--disable-sync`, `--disable-translate`
- Add a third `evaluateOnNewDocument` stealth script in `configurePageDefenses()` covering: Permissions API notifications spoofing (returns `"prompt"` instead of `"denied"`), non-empty `navigator.plugins` array for headless Chrome, and populated `navigator.languages`

## Test plan

- [x] `tests/stealth/stealth-defenses.test.ts` — new test file verifying removed flags are absent from launcher source, `--disable-blink-features=AutomationControlled` is present, and `evaluateOnNewDocument` is called 3 times with functions
- [x] All 1439 existing tests pass (`npm run build && npm test`)

### DIFF
--- a/src/cdp/client.ts
+++ b/src/cdp/client.ts
@@ -1129,7 +1129,10 @@ export class CDPClient {
         });
       }
 
-      // 3. Ensure plugins array is non-empty (headless Chrome has 0 plugins)
+      // 3. Ensure plugins array is non-empty (headless Chrome has 0 plugins).
+      // Individual entries are plain objects, not Plugin instances — sophisticated
+      // detectors using instanceof Plugin will see through this. Turnstile and
+      // most anti-bot systems only check plugins.length > 0.
       if (navigator.plugins.length === 0) {
         Object.defineProperty(navigator, 'plugins', {
           get: () => {
@@ -1157,6 +1160,7 @@ export class CDPClient {
       if (!navigator.languages || navigator.languages.length === 0) {
         Object.defineProperty(navigator, 'languages', {
           get: () => ['en-US', 'en'],
+          configurable: true,
         });
       }
     }).catch(() => {});


shaun0927 (commented):
## PR #265: feat: enhance stealth defenses against anti-bot detection (#257)

No P0 issues found.

- [ ] **[P1]** `navigator.plugins` override missing `configurable: true` — can throw `TypeError` on re-definition — `src/cdp/client.ts:~1133` (Confidence: 85/100)
  - Impact: The `Object.defineProperty(navigator, 'plugins', { get: ... })` call omits `configurable: true` (defaults to `false`). If Puppeteer internals, another `evaluateOnNewDocument` script, or page JavaScript attempts to redefine `navigator.plugins`, it throws `TypeError: Cannot redefine property: plugins`. This silently breaks the page and could cause hard-to-diagnose failures.
  - Fix: Add `configurable: true` to the property descriptor for `navigator.plugins`.

- [ ] **[P1]** Plugin entries are plain objects — fails `instanceof Plugin` checks used by sophisticated bot detectors — `src/cdp/client.ts:~1139-1147` (Confidence: 70/100)
  - Impact: Anti-bot scripts that check `navigator.plugins[0] instanceof Plugin` or iterate `MimeTypeArray` sub-objects will detect the spoofed plugins as fake, potentially making detection *easier* than leaving the array empty. The entries lack `Plugin.prototype` and have no `MimeType` sub-entries.
  - Fix: Either (a) create plugin entries via `Object.create(Plugin.prototype)` and populate `MimeType` entries, or (b) add a code comment documenting this known limitation and why it is acceptable for the current threat model (Turnstile specifically does not do `instanceof` checks).

- [ ] **[P1]** Misleading comment: "Hide chrome.csi and chrome.loadTimes" but code only patches `chrome.runtime` — `src/cdp/client.ts:~1102` (Confidence: 90/100)
  - Impact: The comment says it hides `chrome.csi` and `chrome.loadTimes` but neither property is touched. Misleading comments cause incorrect assumptions during future maintenance and code review.
  - Fix: Change comment to: `// 1. Ensure chrome.runtime exists (real Chrome always exposes it; automation sometimes omits it)`

- [ ] **[P1]** Test for `evaluateOnNewDocument` call count is tautological — mock calls itself 3 times — `tests/stealth/stealth-defenses.test.ts:140-159` (Confidence: 95/100)
  - Impact: The `acquireMockPage` function manually calls `mockPage.evaluateOnNewDocument()` 3 times inside the `createPage` mock, then the test asserts it was called 3 times. This tests the mock, not the production code. `CDPClient` is fully mocked out so the real `configurePageDefenses()` is never exercised. If someone later adds or removes an `evaluateOnNewDocument` call in production code, this test will still pass.
  - Fix: Either (a) use a source-level assertion (like the flag tests do with `fs.readFileSync`) to count `evaluateOnNewDocument(` occurrences in `client.ts`, or (b) import the real `CDPClient.configurePageDefenses` and pass it a mock `Page` object.

- [ ] **[P2]** Unused `evalCallCount` parameter in `acquireMockPage` — `tests/stealth/stealth-defenses.test.ts:140`
  - Suggestion: Remove the unused parameter or wire it up to control mock behavior.

- [ ] **[P2]** Removing `--disable-extensions` may cause real Chrome profiles to load extensions, changing behavior — `src/chrome/launcher.ts:~460`
  - Suggestion: This flag was previously only applied to non-real profiles (`profileType !== 'real'`), so impact is limited. However, temp/managed profiles could still load any user-installed extensions if Chrome's user-data-dir contains them. Consider documenting this behavioral change in release notes.

- [ ] **[P2]** Removing `--disable-ipc-flooding-protection` from the general flags section (all profiles) is a stability trade-off — `src/chrome/launcher.ts:440`
  - Suggestion: This flag previously prevented Chrome from throttling IPC-heavy tabs. Removing it may cause automation commands to be throttled under heavy load. Worth documenting the trade-off.

### Greptile Findings

Cross-referenced with 4 Greptile comments — all valid:
1. **Tautological test** — covered by P1 #4 above.
2. **Misleading comment re: chrome.csi/chrome.loadTimes** — covered by P1 #3 above.
3. **navigator.plugins missing configurable + plain objects** — covered by P1 #1 and #2 above.
4. **Unused evalCallCount parameter** — covered by P2 #1 above.

### Summary

| Priority | Count |
|----------|-------|
| P0 | 0 |
| P1 | 4 |
| P2 | 3 |

### Verdict

⚠️ REQUEST_CHANGES — The stealth defense additions are directionally correct, but the `navigator.plugins` override needs `configurable: true` to avoid `TypeError` in production, the misleading comment should be fixed, and the test suite needs to actually exercise production code rather than verifying its own mocks.

shaun0927 (commented):
## Code Review — PR #265 (oc-code-reviewer + Greptile)

### [P1] Tautological test — mock calls evaluateOnNewDocument 3 times directly (Confidence: 98/100)
**File**: `tests/stealth/stealth-defenses.test.ts:147-162`

The `acquireMockPage` helper manually calls `mockPage.evaluateOnNewDocument()` 3 times inside the mock. The assertion `expect(mockPage.evaluateOnNewDocument).toHaveBeenCalledTimes(3)` is guaranteed to pass regardless of what the real implementation does. Nothing in `configurePageDefenses` is exercised.

**Fix**: Test through the real `configurePageDefenses` path — mock a Puppeteer Page object and pass it to a real CDPClient instance.

### [P1] Comment says "chrome.csi/chrome.loadTimes" but code only patches chrome.runtime (Confidence: 97/100)
**File**: `src/cdp/client.ts:1102-1113`

The comment promises to hide `chrome.csi` and `chrome.loadTimes` (actual automation timing signals), but the code only conditionally defines `chrome.runtime.id` when `chrome.runtime` is absent. In real headed Chrome, `chrome.runtime` already exists, making this block a near no-op.

**Fix**: Either actually stub `chrome.csi`/`chrome.loadTimes`, or correct the comment.

### [P1] navigator.plugins override missing `configurable: true` (Confidence: 95/100)
**File**: `src/cdp/client.ts:1130-1149`

Without `configurable: true`, any SPA that tries to redefine `navigator.plugins` will throw `TypeError`. The `.catch()` on `evaluateOnNewDocument` silently swallows this.

Also, plugin entries are plain objects not real `Plugin` instances — `instanceof Plugin` checks will fail.

**Fix**: Add `configurable: true` to the property descriptor.

### [P2] Unused `evalCallCount` parameter in test
**File**: `tests/stealth/stealth-defenses.test.ts`

### [P2] Removing `--disable-extensions` may load unwanted extensions
For non-real profiles, removing `--disable-extensions` means Chrome could load user-installed extensions that consume memory or interfere with automation. Consider adding a comment noting this trade-off.

### Verdict: Fix P1s before merge

🤖 Reviewed with Claude Code (oc-code-reviewer)

Issue #263, opened 2026-03-09
feat: RTK-inspired Response Compression Layer for 40-67% Token Reduction


shaun0927:
## Summary

Inspired by [RTK (Rust Token Killer)](https://github.com/rtk-ai/rtk) and competitive analysis of [Vercel's agent-browser](https://github.com/vercel-labs/agent-browser), this proposal introduces a **Response Compression Layer** into OpenChrome's MCP response pipeline.

The core insight: most tokens consumed by AI agents come not from reasoning, but from **verbose tool output fed back into context**. OpenChrome's tools routinely return 7,500–12,500 tokens per call, of which only 20–35% is actionable. A centralized compression middleware could reduce output tokens by **40–67%** while **maintaining or improving** task accuracy.

### Why This Is Urgent: Competitive Pressure

Vercel's agent-browser proves that a CLI-first approach with minimal output achieves dramatically better token efficiency:

| Metric | agent-browser (CLI) | OpenChrome (MCP) | Gap |
|--------|---------------------|-------------------|-----|
| Session start overhead | **0 tokens** | ~15K–20K tokens (45 tool schemas) | ∞ |
| Page snapshot | **200–400 tokens** | 7,500–12,500 tokens | **30x** |
| Click response | **6 chars** ("Done") | 300–800 chars (JSON) | **50–130x** |
| 10-step automation | **~7,000 tokens** | ~50K–80K tokens | **7–11x** |

**If we don't close this gap, OpenChrome's richer features (HintEngine, Workers, Orchestration, Domain Memory) become irrelevant** — users will choose the tool that costs 1/10th the tokens for common tasks.

### Why Compression Improves Performance (Not Just Cost)

- **"Lost in the Middle" problem** ([Google Research, 2023](https://arxiv.org/abs/2307.03172)): LLMs lose track of information in long contexts. A 12,000-token DOM tree makes it harder to find a button than a 300-token interactive element list.
- **Signal amplification**: Removing noise (layout divs, tracking cookies, duplicate logs) makes actionable info (interactive elements, errors, state changes) more prominent.
- **Delta mode**: Telling the LLM "5 nodes changed" is more accurate than making it mentally diff two 12,000-token snapshots.

---

### How agent-browser Achieves Token Efficiency

agent-browser takes a radically different architectural approach:

```
agent-browser (CLI):
  LLM → Bash("agent-browser click @e2") → Rust CLI → Unix Socket → Playwright → Chrome
[...]

PR #267, opened 2026-03-09
feat: add compression config and verbosity system

shaun0927:
## Summary
- Adds `compression` configuration block to `GlobalConfig` with `enabled`, `level`, `verbosity`, and `trackSavings` options
- Makes `_timing`, `_profile`, and `_hint` metadata injection conditional based on verbosity level (`compact`/`normal`/`verbose`)
- Adds `_compression` response metadata when `trackSavings` is enabled
- Foundation for all subsequent compression PRs (#263)

## Verbosity Matrix

| Field | compact | normal (default) | verbose |
|-------|---------|-------------------|---------|
| `_timing` | Omitted | `durationMs` only | Full |
| `_profile` | Omitted | Included | Included |
| `_hint` | Critical/error only | All | All |

## Test plan
- [ ] Build passes: `npm run build`
- [ ] Tests pass: `npm test`
- [ ] Default behavior unchanged (verbosity defaults to `normal`)
- [ ] `compact` verbosity omits `_timing` and `_profile`
- [ ] `verbose` verbosity includes full timing

Closes part of #263

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -707,18 +707,29 @@ export class MCPServer {
       if (callId) {
         const timing = this.activityTracker!.getCall(callId);
         if (timing?.duration !== undefined) {
-          (errResult as Record<string, unknown>)._timing = {
-            durationMs: timing.duration,
-            startTime: timing.startTime,
-            endTime: timing.endTime,
-          };
+          const verbosityErr = getGlobalConfig().compression?.verbosity ?? 'normal';
+          if (verbosityErr === 'verbose') {
+            (errResult as Record<string, unknown>)._timing = {
+              durationMs: timing.duration,
+              startTime: timing.startTime,
+              endTime: timing.endTime,
+            };
+          } else if (verbosityErr === 'normal') {
+            (errResult as Record<string, unknown>)._timing = {
+              durationMs: timing.duration,
+            };
+          }
+          // compact: skip _timing entirely
         }
       }
 
       // Inject profile state (no warning on error responses)
-      const profileInfoErr = this.buildProfileInfo();
-      if (profileInfoErr) {
-        (errResult as Record<string, unknown>)._profile = profileInfoErr.profile;
+      const verbosityErr = getGlobalConfig().compression?.verbosity ?? 'normal';
+      if (verbosityErr !== 'compact') {
+        const profileInfoErr = this.buildProfileInfo();
+        if (profileInfoErr) {
+          (errResult as Record<string, unknown>)._profile = profileInfoErr.profile;
+        }
       }
 
       // Inject proactive hint for errors into both _hint and content[]
--- a/src/types/mcp.ts
+++ b/src/types/mcp.ts
@@ -73,4 +73,4 @@ export const MCPErrorCodes = {
 } as const;
 
 /** LLM-side override for compression level on individual tool calls */
-export type CompressionOverride = 'none' | 'light' | 'aggressive' | 'delta';
+export type CompressionOverride = 'none' | 'light' | 'aggressive';


shaun0927 (commented):
## PR #267: feat: add compression config and verbosity system

No P0 issues found.

- [ ] **[P1]** Compact verbosity suppresses `_hint` for `error`-severity hints — only `critical` passes through — `src/mcp-server.ts:~660-662` (Confidence: 85/100)
  - Impact: The hint filter logic is `verbosity !== 'compact' || hintResult.severity === 'critical'`. The PR description's verbosity matrix says compact mode shows "Critical/error only" hints, but the code only passes `critical` — `error` severity hints are suppressed. If the hint engine ever emits `error`-severity hints (the engine currently uses `critical` and `warning`, but the type is not constrained), they would be silently dropped in compact mode.
  - Fix: Either (a) change condition to `hintResult.severity === 'critical' || hintResult.severity === 'error'`, or (b) update the PR description matrix to say "Critical only" to match the actual implementation.

- [ ] **[P1]** Error response path is not covered by verbosity — still injects full `_timing` — `src/mcp-server.ts:~684-692` (Confidence: 90/100)
  - Impact: The verbosity-conditional `_timing` logic only applies to the success path (line ~619). The error/catch path at line ~684 still unconditionally injects the full `_timing` object with `durationMs`, `startTime`, and `endTime` regardless of verbosity setting. This inconsistency means compact mode still gets full timing on errors, defeating the token-saving purpose.
  - Fix: Apply the same verbosity filter to the error path's `_timing` injection.

- [ ] **[P1]** `CompressionOverride` type includes `'delta'` which is not in `GlobalConfig.compression.level` union — `src/types/mcp.ts:76` (Confidence: 90/100)
  - Impact: `GlobalConfig.compression.level` is typed as `'none' | 'light' | 'aggressive'` but `CompressionOverride` adds `'delta'`. Any code that assigns a `CompressionOverride` to a config level (or compares them) will have a type mismatch. Since `CompressionOverride` is exported as a public type, consumers may attempt to use `'delta'` and get runtime inconsistency.
  - Fix: Either add `'delta'` to the `GlobalConfig` level union, or remove it from `CompressionOverride` until delta compression is implemented.

- [ ] **[P2]** Constants in `defaults.ts` are dead code — not imported anywhere — `src/config/defaults.ts:128-136`
  - Suggestion: `DEFAULT_COMPRESSION_LEVEL`, `DEFAULT_VERBOSITY`, and `COMPRESSION_MIN_BYTES` are exported but never imported. `global.ts` uses hardcoded string literals (`'light'`, `'normal'`). Either import and use these constants in `global.ts` for consistency, or defer adding them until they are consumed.

- [ ] **[P2]** `_compression` metadata records config, not actual savings — name `trackSavings` is misleading — `src/mcp-server.ts:~682-686`
  - Suggestion: The `trackSavings` flag implies measured token savings will be reported, but the `_compression` block only records `level` and `verbosity` (configuration echo). Consider renaming to `trackConfig` or `includeCompressionMeta`, or add a comment noting that actual savings tracking is planned for a follow-up.

- [ ] **[P2]** No tests for the verbosity system — `src/mcp-server.ts`
  - Suggestion: This PR changes the response pipeline (success and hint injection paths) but adds no tests. Given that this is the foundation for all subsequent compression PRs (#263), adding unit tests for the three verbosity levels (compact/normal/verbose) would prevent regressions. At minimum, test that compact mode omits `_timing` and `_profile`, and that critical hints still pass through.

- [ ] **[P2]** `compression.tools` per-tool override config is declared but never read — `src/config/global.ts:73-75`
  - Suggestion: The `tools?: Record<string, { level?: ... }>` config field is defined in the interface but no code reads per-tool overrides. Consider removing it from the interface until the follow-up PR implements tool-level compression, to avoid misleading users who configure it expecting it to work.

### Greptile Findings

Cross-referenced with 3 Greptile comments — all valid:
1. **Dead constants in defaults.ts** — covered by P2 #1 above.
2. **`CompressionOverride` includes `'delta'` not in config union** — covered by P1 #3 above.
3. **`_compression` records config not savings** — covered by P2 #2 above.

### Summary

| Priority | Count |
|----------|-------|
| P0 | 0 |
| P1 | 3 |
| P2 | 4 |

### Verdict

⚠️ REQUEST_CHANGES — The error path's `_timing` injection is inconsistent with the success path's verbosity filter, the hint severity filter does not match the documented matrix, and the `CompressionOverride` type mismatch should be resolved before this becomes the foundation for the compression PR chain (#263).

PR #268, opened 2026-03-09
feat: DOM sibling dedup and container collapse for 30-50% token reduction

shaun0927:
## Summary
Implements two DOM compression strategies in `dom-serializer.ts` for issue #263:

- **Sibling Deduplication**: Collapses runs of 4+ consecutive same-tag siblings (e.g., 50 `<li>` items) into first 3 samples + summary line + last element. Reduces list-heavy pages by 40-70%.
- **Container Collapse**: Flattens single-child container chains (`div > div > section > button`) into single-line prefix notation (`[10]div>[11]div>[12]section> [13] button "Submit"`). Reduces deeply nested React/Next.js output by 20-40%.

Default compression is `light` (auto-enabled). Pass `compression: "none"` for full backward-compatible output.

## Key Design Decisions
- All `backendNodeId` refs preserved for element targeting
- Interactive elements never hidden by sibling dedup
- Container collapse only affects non-interactive containers with no text
- Text nodes between element groups preserved

## Test plan
- [ ] Build passes: `npm run build`
- [ ] Tests pass: `npm test`
- [ ] 50 identical `<li>` siblings → collapsed to ~5 lines
- [ ] 3 siblings → NOT collapsed (below threshold)
- [ ] `div > div > section > button` → single line with prefix
- [ ] `compression: "none"` preserves original behavior exactly
- [ ] Interactive elements always visible

Closes part of #263

### DIFF
--- a/src/dom/dom-serializer.ts
+++ b/src/dom/dom-serializer.ts
@@ -111,6 +111,23 @@ function isInteractive(tagName: string, attrMap: Map<string, string>): boolean {
 }
 
 /**
+ * Check if a DOM node or any descendant contains interactive elements.
+ * Prevents sibling dedup from collapsing groups with clickable elements.
+ */
+function containsInteractive(node: DOMNode): boolean {
+  if (node.nodeType !== NODE_TYPE_ELEMENT) return false;
+  const tag = (node.localName || node.nodeName).toLowerCase();
+  const attrMap = parseAttributes(node.attributes);
+  if (isInteractive(tag, attrMap)) return true;
+  if (node.children) {
+    for (const child of node.children) {
+      if (containsInteractive(child)) return true;
+    }
+  }
+  return false;
+}
+
+/**
  * Get direct text content from immediate text node children (not deep)
  */
 function getDirectTextContent(node: DOMNode): string {
@@ -372,7 +389,11 @@ function serializeNode(
     for (const group of groups) {
       if (ctx.truncated) return;
 
-      if (group.nodes.length >= threshold) {
+      // Skip dedup for groups containing interactive elements to avoid
+      // hiding clickable buttons/links/inputs from the LLM
+      const groupHasInteractive = group.nodes.some(n => containsInteractive(n));
+
+      if (group.nodes.length >= threshold && !groupHasInteractive) {
         // Emit first SIBLING_SAMPLE_COUNT with full detail
         const samples = group.nodes.slice(0, SIBLING_SAMPLE_COUNT);
         for (const sampleNode of samples) {


shaun0927 (commented):
## Code Review — PR #268

### Greptile Findings (validated)

**P0: Interactive elements silently dropped by sibling deduplication**
`groupSiblings()` does not check for interactive elements before deduplicating. Buttons, links, and inputs can be collapsed into `×N` summaries, making them unreachable by `find`/`click` tools.

**P0: Container collapse bypasses `interactiveOnly` filter**
The container chain collapse returns early (line 332) without reaching the `interactiveOnly` guard. When `interactiveOnly: true`, collapsed container children are still emitted.

**P1: Dead code loop**
A loop iterates over every child with an empty `if` body — only a comment. Should be removed.

**P1: `backendNodeId` range notation misleading**
`[firstRef-lastRef] tag ×N` implies contiguous ID range but IDs may not be sequential.

**P1: `maxDepth` broken for collapsed chain content**
Leaf children recurse at `depth + 1`, but actual DOM depth is `depth + chain.length`. Deep children escape the depth limit.

### Recommendation
**P0s must be fixed** — interactive element loss breaks core tool functionality. This is a correctness bug that will cause `find` and `click` to fail silently on pages with repeated button/link patterns.

🤖 Reviewed with Claude Code

shaun0927 (commented):
## PR #268: feat: DOM sibling dedup and container collapse for 30-50% token reduction

- [ ] **[P0]** Interactive elements silently dropped by sibling deduplication — `src/dom/dom-serializer.ts:groupConsecutiveSiblings` (Confidence: 95/100)
  - Impact: `groupConsecutiveSiblings` groups by tag name only with no interactivity check. A toolbar with 6 `<button>` elements would show buttons 1, 2, 3 and 6, silently dropping buttons 4 and 5. An LLM agent would have no way to know those buttons exist or interact with them. This defeats the core purpose of the DOM serializer for automation — the PR description explicitly claims "Interactive elements never hidden by sibling dedup" but the code does not enforce this.
  - Fix: Split groups at interactive element boundaries. Inside `groupConsecutiveSiblings`, check `isInteractive(tag, parseAttributes(child.attributes))` and if true, flush the current group and emit the interactive element as a single-node group. Greptile identified this same issue with a concrete fix.

- [ ] **[P1]** Container collapse bypasses `interactiveOnly` filter — `src/dom/dom-serializer.ts:~line 294` (Confidence: 85/100)
  - Fix: When `ctx.interactiveOnly` is true and the container chain collapse fires, it emits non-interactive container context via the chain prefix line, then returns early before the `interactiveOnly` guard. Add a check: if `ctx.interactiveOnly` is true, skip the container collapse path entirely and let normal recursion handle the leaf.
- [ ] **[P1]** `maxDepth` limiting broken for content inside collapsed chains — `src/dom/dom-serializer.ts:~line 326` (Confidence: 90/100)
  - Fix: After collapsing a chain of N containers, the leaf's children are recursed at `depth + 1` but should be at `depth + chain.length + 1` (the leaf's real DOM depth). This allows content well beyond the configured `maxDepth` to be emitted. Change to `serializeNode(child, depth + chain.length + 1, ctx)`.

- [ ] **[P2]** Dead code: empty loop over text nodes — `src/dom/dom-serializer.ts:~line 410`
  - Suggestion: The `for (const child of node.children)` loop at the end of the compression branch has an empty `if` body with only a comment. Remove the entire loop — it does nothing at runtime and adds confusion. Greptile flagged the same issue.
- [ ] **[P2]** `backendNodeId` range notation `[firstRef-lastRef]` is misleading — `src/dom/dom-serializer.ts:~line 385`
  - Suggestion: CDP `backendNodeId` values are not sequential. `[1234-5678] li x50` implies a contiguous range. Consider `[first:1234 ... last:5678]` or similar explicit notation to avoid LLM misinterpretation.
- [ ] **[P2]** No unit tests for the new compression logic
  - Suggestion: Add tests verifying: (a) sibling dedup activates at threshold, (b) interactive elements are not collapsed, (c) container collapse produces correct prefix notation, (d) `compression: "none"` is backward-compatible.

### Greptile Findings
- **Interactive elements dropped by sibling dedup**: Cross-referenced and confirmed. Greptile provided a concrete fix — agree with the approach.
- **Container collapse bypasses interactiveOnly**: Cross-referenced and confirmed.
- **Dead code loop**: Cross-referenced and confirmed.
- **Misleading backendNodeId range**: Cross-referenced and confirmed.
- **maxDepth broken for collapsed chains**: Cross-referenced and confirmed.

### Summary
| Priority | Count |
|----------|-------|
| P0 | 1 |
| P1 | 2 |
| P2 | 3 |

### Verdict
🚫 BLOCK — The P0 (interactive elements silently dropped) is a functional correctness issue that would cause LLM agents to miss actionable UI elements. Must be fixed before merge.

PR #271, opened 2026-03-09
refactor: extract shared element discovery module (#262)

shaun0927:
## Summary

- **Extract shared element discovery module** (`src/utils/element-discovery.ts`) from 5 tool files, eliminating ~1000 lines of duplicated element search and CDP resolution logic
- **Unify element search pattern**: all tools now use `discoverElements()` for in-page search + batched CDP `backendNodeId` resolution
- **Upgrade fill_form CDP resolution** from per-element O(n) round-trips to batched O(1) via shared `resolveBackendNodeIds()`
- **Prerequisite for Shadow DOM support** (#262): shadow-aware traversal can now be added in a single location

### New shared utilities
| Function | Purpose |
|----------|---------|
| `discoverElements()` | In-page interactive element search + batched CDP resolution |
| `discoverFormFields()` | Form-specific field discovery with label detection |
| `resolveBackendNodeIds()` | Batched CDP backendNodeId resolution |
| `getTaggedElementRect()` | Post-scroll element re-positioning |
| `cleanupTags()` | DOM tag cleanup after interaction |

### Files changed
| File | Change |
|------|--------|
| `src/utils/element-discovery.ts` | **NEW** — shared module (468 lines) |
| `src/tools/find.ts` | Refactored: -270 lines |
| `src/tools/click-element.ts` | Refactored: -200 lines |
| `src/tools/interact.ts` | Refactored: -195 lines |
| `src/tools/wait-and-click.ts` | Refactored: -175 lines |
| `src/tools/fill-form.ts` | Refactored: -110 lines |
| `tests/tools/find.test.ts` | Updated CDP mock order |

**Net: -441 lines** (579 added, 1020 removed)

## Test plan

- [x] `npm run build` — zero TypeScript errors
- [x] `npm test` — 75 suites, 1429 tests, all passing
- [x] No regressions in existing find, click_element, interact, wait_and_click, fill_form behavior
- [ ] Manual smoke test on a live page (find → click → interact flow)

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,12 @@ Cookies and localStorage are saved atomically every 30 seconds and restored on s
 
 ---
 
+## Anti-Bot & Turnstile Support
+
+OpenChrome includes built-in defenses against Cloudflare Turnstile and similar anti-bot systems, with additional stealth navigation planned for a future release. See [Turnstile Guide](docs/turnstile-guide.md) for details.
+
+---
+
 ## Benchmarks
 
 Measure token efficiency and parallel performance:
--- a/docs/turnstile-guide.md
+++ b/docs/turnstile-guide.md
@@ -1,0 +1,104 @@
+# Cloudflare Turnstile & Anti-Bot Bypass Guide
+
+## Problem
+
+Cloudflare Turnstile detects browser automation through multiple signals:
+- **CDP fingerprinting**: `Runtime.enable` serialization artifacts reveal CDP connection
+- **navigator.webdriver**: Set to `true` when Chrome DevTools Protocol is active
+- **Chrome launch flags**: Automation-specific flags like `--enable-automation`
+- **screenX/screenY exploit**: CDP mouse events in cross-origin iframes produce incorrect coordinates
+- **Behavioral analysis**: Mouse movement patterns, timing, IP reputation
+
+## OpenChrome's Built-in Defenses
+
+### Automatic Defenses (always active)
+- `--disable-blink-features=AutomationControlled` suppresses `navigator.webdriver`
+- Known automation flags removed from Chrome launch arguments
+- Cookie bridge from authenticated Chrome profile via `useDefaultContext: true`
+
+> **Note**: `--disable-blink-features=AutomationControlled` is not applied when using `chrome-headless-shell`. Always use headed Chrome for Turnstile-protected pages.
+
+### Stealth Navigation Mode (Planned)
+
+> **Status**: This feature is not yet available. It is planned for a future release. See [PR #266](https://github.com/shaun0927/openchrome/pull/266) for implementation progress.
+
+Once available, the `stealth` parameter on the `navigate` tool will open tabs via Chrome's HTTP debug API **without attaching CDP**, allowing Turnstile challenges to complete without detecting automation:
+
+```json
+{
+  "tool": "navigate",
+  "args": {
+    "url": "https://example.com/protected-page",
+    "stealth": true,
+    "stealthSettleMs": 8000
+  }
+}
+```
+
+During the settle period:
+- No `Runtime.enable` is sent
+- No CDP WebSocket is connected to the tab
+- Turnstile's JavaScript sees a normal, unautomated browser
+- The `cf_clearance` cookie is set in Chrome's cookie store
+
+After the settle period, CDP attaches and normal automation resumes. The `cf_clearance` cookie persists because it's bound to the TLS fingerprint and User-Agent (not the CDP session).
+
+**Planned Parameters:**
+- `stealth: true` — Enable CDP-free navigation
+- `stealthSettleMs` — Wait time before CDP attach (default: 5000ms, range: 1000-30000ms)
+
+### Attach Mode Workaround (Available Now)
+If you need to bypass Turnstile today (before stealth mode is available):
+
+1. Start Chrome manually with remote debugging:
+   ```bash
+   # macOS
+   "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" --remote-debugging-port=9222
+
+   # Linux
+   google-chrome --remote-debugging-port=9222
+
+   # Windows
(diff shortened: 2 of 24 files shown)


shaun0927 (commented):
## PR #271: refactor: extract shared element discovery module (#262)

_None identified._

- [ ] **[P1]** Stale `DISCOVERY_TAG` properties corrupt `resolveBackendNodeIds` across sequential tool calls — `src/utils/element-discovery.ts:discoverElements + resolveBackendNodeIds` (Confidence: 85/100)
  - Fix: All tools now share `DISCOVERY_TAG = '__elDiscIdx'`, but `cleanupTags` is **only** called in `wait-and-click.ts` (line 152). When tools run sequentially on the same page (e.g., `find` then `click_element` — a very common pattern), stale `__elDiscIdx` properties from the first call remain on DOM elements. In `resolveBackendNodeIds`, the CDP expression collects **all** elements where `el.__elDiscIdx !== undefined`, including stale elements from prior calls. If a stale element shares the same index as a current element and appears first in document order, the stale element's `backendNodeId` overwrites the current element's slot — causing scroll-into-view and click to operate on the **wrong element**. Add `cleanupTags(page, DISCOVERY_TAG)` after each discovery-and-act sequence in `click-element.ts` (after the click) and `interact.ts` (after the interaction), matching what `wait-and-click.ts` already does.
- [ ] **[P1]** `discoverFormFields` missing element deduplication — `src/utils/element-discovery.ts:discoverFormFields` (Confidence: 90/100)
  - Fix: Unlike `discoverElements` (which uses a `seen: Set<Element>`), `discoverFormFields` has no deduplication. A `<select role="combobox">` (common in Material UI, Ant Design) matches both `'select'` and `'[role="combobox"]'` selectors — it gets added to `fields[]` twice with different `FORM_FIELD_TAG` indices. The DOM property is overwritten to the last index, but `formFields.indexOf(bestMatch)` returns the first entry's index. In `fill-form.ts` lines 189 and 207, the in-page evaluations search for `el[FORM_FIELD_TAG] === firstIndex`, but the element's actual tag is `secondIndex` — the element is **not found**, silently skipping the field or toggling checkbox state incorrectly. Add a `seen` Set identical to the one in `discoverElements`.

- [ ] **[P2]** `find.ts` does not call `cleanupTags` — `src/tools/find.ts`
  - Suggestion: `find` is a read-only tool, so stale tags are less dangerous than for mutation tools. However, since `find` is often the first step before `click_element`, its stale tags directly cause the P1 issue above. Consider adding cleanup in `find.ts` as well, or better yet, call `cleanupTags` at the **start** of `discoverElements` to ensure a clean slate.
- [ ] **[P2]** Test mock ordering change is unexplained — `tests/tools/find.test.ts:line 1 change`
  - Suggestion: The test diff shows a single line change (line 1 → 1 in the diff). Document what CDP mock call ordering changed and why, to avoid confusion for future maintainers.
- [ ] **[P2]** `_origIdx` pattern is fragile and pollutes the type system — multiple files
  - Suggestion: The `(FoundElement & { _origIdx: number })` intersection type is repeated across 4 tool files. Consider adding `_origIdx` as an optional field on `FoundElement` or creating a `ScoredDiscoveryResult` type in `element-discovery.ts` that wraps the score and original index together.

### Greptile Findings
- **Stale DISCOVERY_TAG corruption**: Cross-referenced and confirmed. Greptile identified the exact same cross-call corruption mechanism with a concrete example.
- **discoverFormFields missing deduplication**: Cross-referenced and confirmed. Greptile provided a detailed scenario with `<select role="combobox">` and suggested the `seen` Set fix.

### Summary
| Priority | Count |
|----------|-------|
| P0 | 0 |
| P1 | 2 |
| P2 | 3 |

### Verdict
⚠️ REQUEST_CHANGES — The refactoring is well-structured and achieves significant deduplication (-441 net lines). However, the shared `DISCOVERY_TAG` without consistent cleanup creates a cross-tool corruption vector that did not exist pre-refactoring (each tool previously used its own tag property). The `discoverFormFields` dedup gap can silently skip form fields. Both P1s should be fixed before merge.

shaun0927 (commented):
## Code Review — PR #271 (oc-code-reviewer + Greptile)

### [P1] Stale DISCOVERY_TAG expando properties cause index collision on repeated in-page calls (Confidence: 85/100)
**File**: `src/utils/element-discovery.ts:263`, `src/tools/click-element.ts:108`, `src/tools/interact.ts:115`

Greptile's finding is **directionally correct but the severity is P1, not P0**. When the same SPA page is queried consecutively by different tools (e.g., `find` → `click_element`), stale `__elDiscIdx` tags from the first call can collide with indices from the second call. The `index >= results.length` guard prevents out-of-bounds writes, but if two DOM elements carry the same index, `DOM.describeNode` may resolve the wrong element.

Navigation between calls clears expandos (new document), so this mainly affects SPA workflows without page transitions.

**Fix**: Add `cleanupTags(page, DISCOVERY_TAG)` at the start of each tool handler (or at least at the end). Currently only `wait-and-click.ts` calls cleanup.

### [P1] discoverFormFields missing deduplication — causes wrong element mutation (Confidence: 90/100)
**File**: `src/utils/element-discovery.ts:380-420`

Unlike `discoverElements` (which uses a `seen: Set<Element>`), `discoverFormFields` has no dedup. A `<select role="combobox">` is collected twice with different `__formFieldIdx` values. In `fill-form.ts`, `formFields.indexOf(bestMatch)` finds the first occurrence, but the DOM element's tag was last written with the second index → TreeWalker lookup finds the wrong element.

**Fix**: Add `const seen = new Set<Element>()` guard in `discoverFormFields`, same pattern as `discoverElements`.

### [P2] Good refactoring overall
The extraction of shared element discovery logic is well-structured. The scoring algorithm, tag-based indexing, and backend node resolution are properly abstracted.

### Verdict: Fix P1s (cleanup + dedup) before merge

🤖 Reviewed with Claude Code (oc-code-reviewer)

PR #274, opened 2026-03-09
feat: incremental delta responses for read_page (60-80% reduction on repeats)

shaun0927:
## Summary
Adds DOM snapshot caching and delta responses for `read_page` (#263):

### New: `src/compression/snapshot-store.ts`
- LRU cache (50 entries, 30s TTL) storing serialized DOM per `(sessionId, tabId)`
- Line-based diff computation between cached and current DOM
- Singleton pattern (matches `AdaptiveScreenshot`)
- Falls back to full response if >50% nodes changed

### Modified: `src/tools/read-page.ts`
- New `compression` parameter: `none` | `light` | `delta`
- `delta` mode: caches DOM after serialization, returns diff on next call
- Zero overhead when `compression` is not `delta` (opt-in only)
- Only applies to DOM mode (not AX or CSS)

### Delta Output Format
```
[DOM Delta — 5 of 1,023 nodes changed]

Added (2):
  + [42] div.notification "New message received"
  + [43] span "Saved"

Removed (1):
  - [55] div.loading

Unchanged: 1,020 nodes (99.7%)
```

## Design Decisions
- Distinct from `withDomDelta` (action-level MutationObserver) — this is read-level structural diff
- 30s TTL prevents stale cache
- 50% change threshold auto-falls back to full (page fundamentally changed)
- Opt-in via `compression: "delta"` — no default behavior change

## Test plan
- [x] Build passes: `npm run build`
- [x] Tests pass: `npm test` (75 suites, 1429 tests)
- [ ] First `read_page(compression: "delta")` → full DOM + cached
- [ ] Second call (same page) → delta with added/removed counts
- [ ] URL change → full DOM (cache invalidated)
- [ ] >50% nodes changed → full DOM fallback
- [ ] `compression: "none"` or omitted → exact current behavior
- [ ] Cache TTL (30s) → stale entries evicted

Closes part of #263

### DIFF
--- a/src/compression/snapshot-store.ts
+++ b/src/compression/snapshot-store.ts
@@ -114,28 +114,25 @@ export class SnapshotStore {
     const prevLines = previous.content.split('\n');
     const currLines = currentContent.split('\n');
 
-    // Build a set of previous lines for fast lookup
+    // Build a set of previous lines for fast lookup (preserve indentation for structural comparison)
     const prevSet = new Map<string, number>();
     for (const line of prevLines) {
-      const trimmed = line.trim();
-      if (trimmed) prevSet.set(trimmed, (prevSet.get(trimmed) || 0) + 1);
+      if (line.trim()) prevSet.set(line, (prevSet.get(line) || 0) + 1);
     }
 
     const currSet = new Map<string, number>();
     for (const line of currLines) {
-      const trimmed = line.trim();
-      if (trimmed) currSet.set(trimmed, (currSet.get(trimmed) || 0) + 1);
+      if (line.trim()) currSet.set(line, (currSet.get(line) || 0) + 1);
     }
 
     // Find added lines (in current but not in previous)
     const added: string[] = [];
     const tempPrev = new Map(prevSet);
     for (const line of currLines) {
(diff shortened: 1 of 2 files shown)


shaun0927 (commented):
## Code Review — PR #274

### Greptile Findings (validated)

**P1: `light` compression mode is a silent no-op**
`compression: "light"` is listed as a valid enum value but the code only checks `if (compression === 'delta')`. Light mode silently returns uncompressed data.

**P1: Trimming lines discards structural information**
All lines compared after `.trim()` strips indentation. Since indentation encodes tree depth in the DOM serializer, two structurally different DOMs with the same text content will produce false "unchanged" reports.

### Additional Observations
- The snapshot store is a good concept for reducing repeated `read_page` token usage
- Delta computation is sound for large page re-reads
- Memory management (Map-based store) should have TTL/eviction to prevent unbounded growth

### Recommendation
Fix the `light` mode no-op — either implement it or remove it from the enum. The `.trim()` indentation issue is subtle but can cause missed updates in deeply nested SPA content.

🤖 Reviewed with Claude Code

PR #283, opened 2026-03-09
feat: extend supporting utilities with shadow DOM traversal (#262)

shaun0927:
## Summary

Phase 5 of the shadow DOM support plan (#262). Extends three supporting utility modules to traverse open shadow roots:

- **dom-delta.ts**: `MutationObserver` now attaches separate observers per open shadow root, detects new shadow roots on dynamically added nodes, and disconnects all shadow observers on collection/cleanup
- **visual-summary.ts**: Replaces all `document.querySelectorAll` calls with inline `deepQSA` that pierces open shadow roots; follows the `shadowRoot.activeElement` chain to find the actual focused element inside shadow DOM
- **page-diagnostics.ts**: Adds recursive `deepElementCount` that traverses open shadow roots for accurate total element counts

### Changes

| File | Change |
|------|--------|
| `src/utils/dom-delta.ts` | Shadow-aware MutationObserver with per-shadow-root observers |
| `src/utils/visual-summary.ts` | Deep QSA + active element shadow chain traversal |
| `src/utils/page-diagnostics.ts` | Recursive deep element counting through shadow roots |

### Stacked PR

This PR is stacked on Phase 4 (#282). The full phase chain:
- Phase 1: DOM serializer shadow traversal (#270)
- Phase 2: CDP-based shadow-aware element discovery (#275)
- Phase 3: JS deep traversal for open shadow roots (#278)
- Phase 4: Tool-specific shadow DOM enhancements (#282)
- **Phase 5: Supporting utilities shadow DOM extension (this PR)**

## Test plan

- [x] `npm run build` compiles without errors
- [x] `npm test` — all 1429 tests pass
- [ ] Verify MutationObserver captures mutations inside shadow DOM elements
- [ ] Verify visual summary reports headings/forms/active elements inside shadow roots
- [ ] Verify page diagnostics counts elements inside shadow roots

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #284, opened 2026-03-09
test: add comprehensive shadow DOM test suite (#262)

shaun0927:
## Summary

Phase 6 (final phase) of the shadow DOM support plan (#262). Adds comprehensive test coverage for all shadow DOM functionality introduced in Phases 1-5.

**82 new tests** across 4 test files, all passing with zero regressions on the existing 1429 tests.

### Test Files

| File | Tests | Coverage |
|------|-------|----------|
| `shadow-dom-serializer.test.ts` | 22 | DOM serializer shadow root rendering (open/closed/user-agent, nested, depth limiting, interactive filter) |
| `shadow-dom-discovery.test.ts` | 30 | Shadow root tree walking, CSS queries in shadow roots, element discovery with scoring |
| `shadow-dom-query.test.ts` | 17 | query_dom CSS/XPath shadow piercing, pierceShadow parameter, fallback behavior |
| `shadow-dom-utils.test.ts` | 13 | dom-delta MutationObserver, visual-summary deepQSA, page-diagnostics element count |

### Key Test Scenarios

- Open, closed, and user-agent shadow root rendering with correct boundary markers
- Nested shadow roots (shadow root inside shadow root)
- Depth limiting across shadow root boundaries
- Interactive-only filter inside shadow roots
- CSS/XPath queries piercing shadow DOM via CDP fallback
- `pierceShadow: false` disabling shadow traversal
- MutationObserver crossing shadow root boundaries
- Active element following shadow root chain
- Recursive deep element counting through shadow roots

### Stacked PR

This PR completes the full shadow DOM support chain:
- Phase 1: DOM serializer shadow traversal (#270)
- Phase 2: CDP-based shadow-aware element discovery (#275)
- Phase 3: JS deep traversal for open shadow roots (#278)
- Phase 4: Tool-specific shadow DOM enhancements (#282)
- Phase 5: Supporting utilities shadow DOM extension (#283)
- **Phase 6: Comprehensive test suite (this PR)**

## Test plan

- [x] `npm run build` compiles without errors
- [x] All 82 new shadow DOM tests pass
- [x] All 1429 existing tests pass (zero regressions)
- [x] Total: 1511 tests passing


shaun0927 (commented):
## PR #284: test: add comprehensive shadow DOM test suite (#262)

**Verdict**: COMMENT
**P0**: 0 | **P1**: 3 | **P2**: 4

This PR is actually a multi-phase shadow DOM implementation (Phases 1-6 bundled together from a stacked PR chain) covering DOM serializer, element discovery, JS deep traversal, tool-specific enhancements, supporting utilities, and test coverage. The refactoring of 5 tool files into a shared `element-discovery.ts` module is well-structured. The shadow DOM coverage itself is comprehensive and technically sound. Greptile left an empty review body.

---

### Greptile Findings
Greptile submitted a COMMENTED review with an empty body — no findings to report.

---

### Issues Found

**[P1] `fill-form.ts` checkbox/select re-lookup uses light-DOM-only `TreeWalker` after shadow-aware discovery**
`src/tools/fill-form.ts` (diff lines ~617-652 / `FORM_FIELD_TAG` usage in checkbox and select branches)

`discoverFormFields()` now uses `deepQuerySelectorAll` to tag elements with `FORM_FIELD_TAG` inside open shadow roots. However, the code that later reads back those tagged elements to check `isChecked` or set a select value uses a plain `createTreeWalker(document.body, ...)` — this walker does not pierce shadow roots. Any form field inside a shadow root gets tagged correctly during discovery but is unreachable during the subsequent interaction step, causing the checkbox state check to silently return `false` and the select assignment to silently no-op.

Fix: Replace the inline `createTreeWalker` searches in both the checkbox and select branches with the same `deepWalkElements` pattern already used in `resolveBackendNodeIds` — or better, inject `DEEP_WALK_ELEMENTS_JS` via `Runtime.evaluate` as done in that function.

---

**[P1] `DISCOVERY_TAG` (`__elDiscIdx`) is a shared global property — concurrent tool calls on the same page corrupt each other's indices**
`src/utils/element-discovery.ts:52`

All five tools (`find`, `click_element`, `interact`, `wait_and_click`, `fill_form`) now share a single constant `DISCOVERY_TAG = '__elDiscIdx'`. In a multi-worker workflow where two parallel workers interact with the same tab simultaneously (or a tool polling loop overlaps with a concurrent call), the second `discoverElements` call overwrites DOM property indices set by the first. This was not a risk with the old per-tool property names (`__clickIndex`, `__interactIndex`, `__findIndex`, `__waitClickTarget`).

Fix: Generate a unique tag per discovery call (e.g. `__elDiscIdx_${Date.now()}` or a monotonic counter), or enforce that concurrent tool calls on the same tab are serialized. The same concern applies to `FORM_FIELD_TAG`.

---

**[P1] `wait_and_click` does not call `cleanupTags` on timeout or early error exit**
`src/tools/wait-and-click.ts` (around the timeout exit path)

`cleanupTags(page, DISCOVERY_TAG)` is only called after a successful click. When the polling loop times out (element never found), the function returns an error result without cleaning up `__elDiscIdx` properties left on DOM elements by the last `discoverElements` call. These stale tags persist across subsequent tool calls and can corrupt the next tool's index resolution (amplifying the P1 above).

Fix: Add a `finally` block or explicit cleanup before the timeout-error return path:
```ts
// Before returning the timeout error:
await cleanupTags(page, DISCOVERY_TAG).catch(() => {});
```

---

**[P2] `deepQuerySelectorAll` is re-implemented inline in 4+ separate `page.evaluate` callbacks**
Files: `src/utils/element-discovery.ts`, `src/utils/ref-id-manager.ts`, `src/tools/inspect.ts`, `src/tools/query-dom.ts` (XPath handler)

`DEEP_QUERY_SELECTOR_ALL_JS` exists in `shadow-dom.ts` as an injectable string for `Runtime.evaluate` contexts, but for `page.evaluate` callbacks each file defines its own copy of the same function as a TypeScript inline function. Four separate divergence points. The `ref-id-manager.ts` version also uses `(all[j] as any).shadowRoot` — this `as any` is purely cosmetic since the code runs in the browser, but it flags as inconsistent with the typed versions in other files.

Fix: Export a shared TypeScript function string constant usable in `page.evaluate` closures, or accept the duplication with a comment tying all copies back to a canonical source.

---

**[P2] `shadowCSSFallback` in `query-dom.ts` resolves elements serially**
`src/tools/query-dom.ts:180-219`

The fallback iterates `backendNodeIds` in a sequential `for` loop, making one `DOM.resolveNode` + `Runtime.callFunctionOn` round-trip pair per element. For a selector matching many shadow elements (up to 50), this is up to 100 sequential CDP calls. The existing `resolveBackendNodeIds` utility in `element-discovery.ts` demonstrates the batched pattern.

Fix: Collect all `DOM.resolveNode` + `Runtime.callFunctionOn` promises and `Promise.all` them, as done elsewhere in this PR.

---

**[P2] `dom-delta.ts` error-path observer cleanup is a difficult-to-read inline string**
`src/utils/dom-delta.ts:2030` (in the `withDomDelta` catch block)

The new shadow observer cleanup on action failure is embedded as a single long inline evaluated string that's hard to audit. The main observer disconnect (already a one-liner) is fine, but the shadow observers loop should match the `COLLECT_DELTA_SCRIPT` pattern for consistency and reviewability.

Fix: Extract the combined cleanup expression into a named constant `DISCONNECT_OBSERVERS_SCRIPT` alongside the other script constants.

---

**[P2] `dom-delta.ts` shadow observer walk on injection may be expensive on large pages**
`src/utils/dom-delta.ts` (`walkForShadowRoots` called in `INJECT_OBSERVER_SCRIPT`)

The injection script now calls `walkForShadowRoots(document)` synchronously, which queries `querySelectorAll('*')` on the entire document to find existing shadow roots before the action. On pages with thousands of elements (SPAs, data grids), this adds latency to every `withDomDelta` call, even on pages with no shadow DOM.

Fix: Wrap the initial walk in a fast guard — only walk if any shadow root is likely present (e.g. check `document.querySelector('*:defined')` is not reliable, but a quick `document.querySelectorAll('*').length < 5000` budget check would suffice), or make the shadow observer opt-in rather than always-on.

---

### Summary

The shadow DOM implementation is architecturally sound and the test coverage is thorough. The P1 issues are the most actionable: the `fill-form` light-DOM-only re-lookup after shadow-aware discovery is a functional regression for shadow DOM form fields, the shared `DISCOVERY_TAG` creates a concurrency hazard in multi-worker sessions, and missing cleanup on `wait_and_click` timeout leaves stale DOM state. None of these are blockers for the test PR itself, but they should be addressed before the shadow DOM feature is considered fully production-ready.

Issue #299, opened 2026-03-12
fix: hint engine fails to prevent LLM hang loops — 3 architectural gaps


shaun0927:
## Problem

The hint engine is supposed to detect when the LLM agent is stuck and guide it toward recovery. In practice, **3 architectural gaps** cause the hint engine to be ineffective against the most common hang patterns.

### Evidence: Real hang patterns observed

```
⏺ form_input (ref: "1050", value: "Standard edition") → Error: Runtime.callFunctionOn timed out
⏺ computer (action: "screenshot") → [success, screenshot returned]
⏺ form_input (ref: "1050", value: "Standard edition") → Error: Runtime.callFunctionOn timed out
⏺ computer (action: "screenshot") → [success, screenshot returned]
⏺ form_input (ref: "1050", value: "Standard edition") → Error: Runtime.callFunctionOn timed out
... (hangs indefinitely)
```

The hint engine sees each error individually but never recognizes the repeating pattern because interleaved screenshots reset its detectors.

### Gap 1: Rule coverage holes — 3 error types have NO matching rule

| Error message | Expected rule | Actual match |
|---|---|---|
| `"Target X not found in session Y"` | `error-recovery-1` (`/tab.*not found/`) | **NO MATCH** — message says "Target", not "tab" |
| `"Element is not editable: path"` | Any of 10 rules | **NO MATCH** — no rule covers "not editable" |
| `"Could not get position for ref: ...layout object"` | Any of 10 rules | **NO MATCH** — no rule covers "layout object" |

`error-recovery-1` regex (`/tab[^a-z]*not found/`) doesn't match `buildStaleTargetError()` output format which uses "Target X not found in session Y".

### Gap 2: Interleaved observation tools defeat stuck detection (critical)

The ProgressTracker counts consecutive non-progress calls to detect "stuck" (≥3 errors) and "stalling" (≥3 non-progress). However, the LLM's natural behavior of taking screenshots between retries **resets the counter**:

```
Progress tracker evaluation at call 5 (form_input error):
  consecutiveErrors = 1 (current call)
  → recentCalls[0] = call 4 (screenshot, success)
  → isLikelyProgressCall(screenshot) = true  ← WRONG: screenshot is observation, not progress
  → break → consecutiveErrors = 1 → status: "progressing"
```

The same issue affects `same-tool-error-streak` (priority 250): it requires consecutive calls of the same tool, but screenshots break the streak.

[...]

PR #300, opened 2026-03-12
fix: add missing error-recovery rules and fix tab-not-found regex

shaun0927:
## Summary

- **Fix A**: Add two new error-recovery hint rules for previously uncovered error patterns:
  - `not editable` — fires when `form_input` targets a non-editable element (SVG, div, span), guiding the LLM to use `find` or `javascript_tool`
  - `layout object` / `could not get position` — fires when `computer` tool can't resolve coordinates for hidden/detached elements
- **Fix B**: Broaden the tab-not-found regex to match session-manager's actual error format (`Target X not found in session Y`), which previously didn't match the rule's pattern

These are minimal, targeted additions (~10 lines) to the existing first-match-wins rule chain. No architectural changes.

## Test plan

- [x] `npm run build` passes
- [x] `npm test` — all 66 test suites pass
- [x] Verified regex patterns match real error strings from `form-input.ts` and `computer.ts`
- [ ] Manual verification: trigger "not editable" and "layout object" errors in a live session

Refs: #299

### DIFF
--- a/src/hints/rules/error-recovery.ts
+++ b/src/hints/rules/error-recovery.ts
@@ -7,15 +7,15 @@ import type { HintRule } from '../hint-engine';
 
 const patterns: Array<{ test: RegExp; hint: string }> = [
   {
-    test: /ref[^a-z]*not found|invalid ref|stale ref/i,
+    test: /ref\b.+not found|invalid ref|stale ref/i,
     hint: 'Hint: Refs expire after page changes. Use read_page or find for fresh refs.',
   },
   {
-    test: /tab[^a-z]*not found|target[^a-z]*not found|invalid tab|no such tab/i,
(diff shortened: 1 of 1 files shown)


shaun0927 (commented):
## Code Review

**Verdict: PASS** (after P0 fix applied)

### Issues Found & Fixed
- **P0 (fixed)**: `[^a-z]*` with `/i` flag excluded all letters, causing regex to fail when IDs contain letters. Fixed by replacing with `\b.+` word boundary pattern across all affected regexes (ref, tab, target, selector).

### Verified
- New `not editable` and `layout object` rules match actual error strings from `form-input.ts` and `computer.ts`
- `target\b.+not found` now matches session-manager's `Target ${targetId} not found` format
- `\b` word boundary prevents false positives (e.g., "table" doesn't match "tab")
- All 152 hint tests pass
- Build clean

Issue #344, opened 2026-03-20
feat: Long-Running Session Hardening — Heartbeat & Reconnection Enhancement


shaun0927:
## Summary

Enhance openchrome's heartbeat and reconnection mechanisms to support **1-2+ hour continuous browser automation sessions** without connection drops, state loss, or progressive degradation.

The current implementation already has solid foundations (5s heartbeat, auto-reconnect, circuit breaker). This issue addresses the remaining gaps for truly persistent, long-running browser automation — inspired by Paperclip-style heartbeat orchestration patterns.

## Motivation

Real-world use cases that require long sessions:
- QA testing across hundreds of pages (ralph loop + browser automation)
- Data extraction from complex authenticated web apps
- Multi-step workflow automation (e.g., form submission pipelines)
- Continuous monitoring dashboards

Current limitations for 1h+ sessions:
- Reconnect attempts capped at 3 with flat delay (no backoff)
- Heartbeat interval hardcoded (no env var override)
- Session state is in-memory only (lost on MCP server restart)
- No per-tab renderer health detection
- Storage state watchdog disabled by default

### SC-1: 1-Hour Continuous Operation
- [ ] Complete 60 minutes of continuous browser automation (navigate + interact every 30s) across 3+ different websites without any unrecoverable connection failure

### SC-2: Forced Disconnect Recovery
- [ ] Survive a forced Chrome kill (`kill -9`) and auto-reconnect within 30 seconds
- [ ] Resume operation on surviving/reopened tabs after reconnect

### SC-3: GC Pause Tolerance  
- [ ] Survive Chrome GC pauses of up to 10 seconds without false disconnect detection
- [ ] Exponential backoff prevents premature reconnect exhaustion

### SC-4: Session State Preservation
- [ ] Auth cookies and localStorage persist across Chrome restart (storage watchdog)
- [ ] Tab-to-worker mappings survive MCP server graceful restart

### SC-5: Memory Stability at Scale
- [ ] No unbounded growth in circuit-breaker, strategy-learner, or memory store after 100+ interactions
- [ ] Heap usage delta < 50MB after 1 hour of continuous operation

### SC-6: Per-Tab Health Detection
- [ ] Detect and report a frozen/crashed renderer tab independently of global heartbeat
- [ ] Evict unresponsive tabs without killing the entire session

#### 1.1 Exponential Backoff for Reconnection
- Replace flat `reconnectDelayMs` with `delay * 2^attempt + jitter`
[...]

PR #346, opened 2026-03-20
feat: enable storage state watchdog by default (#344)

shaun0927:
## Summary

- Add `DEFAULT_WATCHDOG_INTERVAL_MS` to `config/defaults.ts`
- Fix bug: `DEFAULT_STORAGE_STATE_RESTORE_TIMEOUT_MS` existed in defaults but was unused — replace hardcoded `10000` in `storage-state-manager.ts`
- Replace hardcoded `30000` watchdog interval with `DEFAULT_WATCHDOG_INTERVAL_MS` in both `storage-state-manager.ts` and `session-manager.ts`
- Support `OPENCHROME_WATCHDOG_INTERVAL_MS` environment variable override
- **Enable storage state persistence by default** (was opt-in via `OC_PERSIST_STORAGE=1`, now opt-out via `OC_PERSIST_STORAGE=0`)

## Motivation

Part of #344 (Long-Running Session Hardening). For 1-2+ hour sessions doing authenticated work, an unplanned Chrome crash previously meant losing all auth cookies. By enabling the storage state watchdog by default, cookies and localStorage are automatically persisted to disk every 30 seconds, allowing recovery after reconnect.

## Changes

| File | Change |
|------|--------|
| `src/config/defaults.ts` | +1 new constant |
| `src/storage-state/storage-state-manager.ts` | Import + use constants (bug fix + consistency) |
| `src/session-manager.ts` | Use constant + env var; flip storage from opt-in to opt-out |
| `tests/storage-state/storage-state-manager.test.ts` | +2 new tests (default interval, restore timeout) |

## Breaking change note

Storage state persistence is now **enabled by default**. Users who explicitly don't want it can set `OC_PERSIST_STORAGE=0`. This is a behavioral change but improves the default experience for long-running sessions.

## Test plan

- [ ] `npx jest tests/storage-state/storage-state-manager.test.ts` — 17/17 pass
- [ ] `npm run build` — clean compilation
- [ ] Verify opt-out: `OC_PERSIST_STORAGE=0 npx openchrome-mcp` skips storage state

### DIFF
--- a/src/config/defaults.ts
+++ b/src/config/defaults.ts
@@ -72,6 +72,11 @@ export const DEFAULT_COOKIE_CONTEXT_TIMEOUT_MS = 5000;
 /** Storage state restore timeout in milliseconds. */
 export const DEFAULT_STORAGE_STATE_RESTORE_TIMEOUT_MS = 10000;
 
+/** Storage state watchdog interval in milliseconds.
+ *  How frequently cookies and localStorage are persisted to disk.
+ *  Override with OPENCHROME_WATCHDOG_INTERVAL_MS environment variable. */
+export const DEFAULT_WATCHDOG_INTERVAL_MS = 30000;
+
 /** createTarget aggregate timeout in milliseconds. Safety net for entire tab creation chain. */
 export const DEFAULT_CREATE_TARGET_TIMEOUT_MS = 60000;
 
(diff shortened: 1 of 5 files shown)


shaun0927 (commented):
## Code Review — PR #346

**Verdict**: APPROVE (after fixes)

### Fixes Applied
- `35daa61`: Fixed `createSnapshot` type contract (Set<string> instead of Map), added save() queueing, improved clear() error handling

### Remaining P2s (follow-up)
- `clear()` uses dynamic `import('fs/promises')` where static import would do
- Storage-on-by-default is a behavioral change — add startup log message
- Test name for `OC_PERSIST_STORAGE=1` is now misleading

P0: 0 (fixed) | P1: 0 (fixed) | P2: 3 (deferred)

Issue #369, opened 2026-03-23
feat: Runtime multi-profile support — run multiple Chrome profiles simultaneously


shaun0927:
## Problem

Currently, the Chrome profile is fixed at MCP server startup via `--profile-directory` or `CHROME_PROFILE_DIRECTORY`. There is no way to switch profiles or use multiple profiles during a session. Users who work with multiple Chrome profiles (e.g., personal account, work account) must restart the server to change profiles.

In real Chrome usage, users routinely open multiple windows with different profiles simultaneously. OpenChrome should support the same pattern.

## Current Architecture

```
MCP Server → ChromeLauncher (singleton) → Chrome Process (single)
                                              └── --profile-directory fixed at launch
```

**What already exists:**
- `ChromePool` (`src/chrome/pool.ts`) — manages up to 5 Chrome instances on different ports (9222, 9223, ...), currently keyed by **origin** to avoid CDP same-origin renderer contention
- `workerId` parameter — supported by `navigate`, `tabs_create`, and other tools; provides isolated browser contexts with separate cookie jars and tab pools
- `SessionManager.createWorker()` — already routes workers to different ChromePool instances when `usePool=true`
- `list_profiles` tool — discovers all local Chrome profiles from `Local State`
- `ProfileManager.resolveProfile()` — handles profile resolution with 5-tier priority

**What's missing:**
- ChromePool has no concept of **profile** — all instances use the same default profile
- No MCP tool parameter to specify which profile a navigation/tab should use
- No mapping from profile → Chrome instance in the pool

### Architecture

```
MCP Server → ChromePool (profile-aware)
               ├── Port 9222: Chrome --profile-directory="Default"     ← personal
               ├── Port 9223: Chrome --profile-directory="Profile 1"   ← work
               └── Port 9224: Chrome --profile-directory="Profile 2"   ← client
```

Each profile runs as a separate Chrome process on a separate debug port, with its own `SingletonLock`. This mirrors how Chrome natively handles multi-profile usage.

### Key Constraint

Same `--user-data-dir` + same `--profile-directory` cannot be opened by two Chrome processes simultaneously (Chrome's SingletonLock). This means: **one Chrome instance per profile**, which is the natural and expected behavior.

### Phase 1: ChromePool profile-awareness

**File: `src/chrome/pool.ts`**

1. Add `profileDirectory` to `PooledInstance`:
   ```ts
   export interface PooledInstance {
     port: number;
     launcher: ChromeLauncher;
[...]

PR #370, opened 2026-03-23
feat(pool): add profile-aware Chrome instance management

shaun0927:
## Summary

- Add `acquireInstanceForProfile()` to `ChromePool` for profile-specific Chrome instance routing — reuses existing instances matching the requested profile or launches a new Chrome process with `--profile-directory`
- Add `ensurePool()` to `SessionManager` for lazy `ChromePool` initialization on first multi-profile request (no config flag needed)
- Thread `profileDirectory` through `createWorker()` → `createTarget()` → `createTargetStealth()` so MCP tools can request specific Chrome profiles
- Update `Worker` and `WorkerCreateOptions` types with `profileDirectory` field

This is the core infrastructure for #369. The MCP tool interface (Phase 2) and UX updates (Phase 4/5) follow in a stacked PR.

## Test plan

- [x] `npm run build` passes
- [x] All 2083 unit tests pass — zero regressions
- [ ] Manual: start server, call `navigate` without `profileDirectory` — existing behavior unchanged
- [ ] Manual: call `navigate` with `profileDirectory: "Profile 1"` — new Chrome instance launches on separate port with correct profile

### DIFF
--- a/src/chrome/pool.ts
+++ b/src/chrome/pool.ts
@@ -18,6 +18,7 @@ export interface PooledInstance {
   origins: Set<string>;  // origins currently using this instance
   tabCount: number;
   isPreExisting: boolean; // was it already running when we found it?
+  profileDirectory?: string; // Chrome profile directory (e.g., "Profile 1", "Default")
 }
 
 const DEFAULT_POOL_CONFIG: ChromePoolConfig = {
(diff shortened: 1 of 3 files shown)


shaun0927 (commented):
## Code Review — P0: 0, P1: 5, P2: 2

### [P1] Browser context created against wrong CDP client before profile instance is connected
Context is created via primary `cdpClient` before profile-specific instance is connected.

### [P1] `releaseInstance` not called on worker cleanup when `profileDirectory` is set without `targetUrl`
`workerPoolOrigin` is undefined, so the guard in `deleteWorkerInternal` is never satisfied.

### [P1] Concurrent `acquireInstanceForProfile` calls can race-launch duplicate Chrome processes
No in-flight deduplication for same profile.

### [P1] `new URL(options.targetUrl)` throws uncaught on malformed URL

### [P1] `getChromePool` singleton ignores config on second call (pre-existing, surfaced by lazy init path)

Fixing these in follow-up commits.

Issue #347, opened 2026-03-20
feat: Self-Healing Browser Automation — 24-Hour Session Architecture


shaun0927:
## Summary

Design and implement a **5-layer self-healing architecture** that enables openchrome to sustain browser automation sessions for 1–24 hours without hangs, state loss, or unrecoverable failures.

This goes beyond Phase 1's parameter tuning (#344) — it adds **new defense layers** that make openchrome a true harness for long-running automation.

## Problem Statement

Current failure modes for long sessions:

| Duration | Failure Risk | Root Cause |
|----------|:--:|-----------|
| 0–30 min | Low (5%) | Transient CDP errors — already handled |
| 30–60 min | Medium (20%) | Chrome GC pauses, session TTL edge cases |
| 1–3 hours | High (40%) | Chrome crash, memory leaks, WebSocket half-open |
| 3–24 hours | Very High (60%) | Process crashes, context compaction loss, resource exhaustion |

Three distinct breakpoints exist in the system:

```
[AI Agent] ←── ① MCP JSON-RPC ──→ [openchrome MCP Server] ←── ② CDP WebSocket ──→ [Chrome]
                                              │
                                          ③ Internal State
                                          (in-memory only)
```

All three must be defended for 24-hour sessions.

## Architecture: 5-Layer Defense-in-Depth

```
┌─────────────────────────────────────────────────────────────────┐
│  LAYER 5: Process Supervisor                                     │
│  PM2 / systemd with exp_backoff_restart_delay                   │
│  Nightly 4am restart as safety valve                             │
│  max_memory_restart: 512M                                        │
├─────────────────────────────────────────────────────────────────┤
│  LAYER 4: Application Watchdog                                   │
│  Event loop block detection (blocked / native-watchdog)          │
│  30s check-in timeout → process.exit(1) → Layer 5 restarts      │
│  Health endpoint on :9229 for external monitoring                │
├─────────────────────────────────────────────────────────────────┤
│  LAYER 3: Chrome Process Supervisor                              │
│  PID watchdog: 10s interval, detect crash → auto-relaunch        │
│  Profile persistence (--restore-last-session + --user-data-dir)  │
│  Stability flags for long sessions (--disable-dev-shm-usage)     │
├─────────────────────────────────────────────────────────────────┤
│  LAYER 2: Session State Persistence                              │
│  targetToWorker → disk (debounced 5s writes)                     │
[...]

PR #379, opened 2026-03-23
feat: activate heavy heartbeat mode and add Chrome stability flags (#347)

shaun0927:
## Summary

- **Heavy heartbeat mode during tool execution**: Switch from `active` to `heavy` (2s interval) when a tool starts executing, matching the Phase 3B adaptive heartbeat spec. Immediately transitions back to `active` on tool completion (both success and error paths) before the 5-minute idle timer fires.
- **Remove dead code in CDPClient**: The idle-check inside `startHeartbeat()` depended on `lastCommandAt` which was never set externally — the condition `lastCommandAt > 0` was never true. Removed that dead branch. `recordCommandActivity()` is retained but marked `@deprecated` for API compatibility.
- **Chrome stability flags**: Add three missing flags for long sessions: `--disable-crash-reporter`, `--disable-session-crashed-bubble`, `--hide-crash-restore-bubble` to suppress crash UI that blocks automation.

## Test plan

- [ ] `npm run build` passes with no TypeScript errors
- [ ] `npm test` — all 2083 tests pass across 104 suites
- [ ] Heartbeat mode transitions: tool start → `heavy`, tool end → `active`, after 5min idle → `idle`
- [ ] Chrome launcher: new flags appear in launched Chrome process args
- [ ] No regressions in CDP reconnect or session management behavior

### DIFF
--- a/src/cdp/client.ts
+++ b/src/cdp/client.ts
@@ -223,14 +223,6 @@ export class CDPClient {
         return;
       }
 
-      // Check for idle transition: no commands for 5 minutes → switch to idle mode
-      if ((this.heartbeatMode === 'active' || this.heartbeatMode === 'heavy')
-          && this.lastCommandAt > 0
-          && now - this.lastCommandAt > 300000) {
-        this.setHeartbeatMode('idle');
-        return; // setHeartbeatMode restarts the timer with new interval
-      }
-
       this.checkConnection();
     }, this.getEffectiveHeartbeatInterval());
   }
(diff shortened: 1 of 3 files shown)


shaun0927 (commented):
**Code Review — PR #379**

Three-part change for Chrome stability in long sessions:

1. **cdp/client.ts**: Removes dead idle-transition code from heartbeat loop; idle transitions now managed by MCPServer via setTimeout. Deprecated annotation added for API compat.
2. **chrome/launcher.ts**: Adds `--disable-crash-reporter`, `--disable-session-crashed-bubble`, `--hide-crash-restore-bubble` flags to suppress crash UI blocking automation.
3. **mcp-server.ts**: Switches tool execution to heavy heartbeat mode (was active), adds post-tool transition back to active mode.

**P0**: 0 | **P1**: 0 | **P2**: 0

LGTM — merging.

Issue #403, opened 2026-03-24
Reliability Guarantee Initiative — Master Tracking Issue


shaun0927:
# Reliability Guarantee Initiative — Master Tracking Issue

> **Philosophy:** Every tool call MUST return a result — success or error — within the timeout. OpenChrome never hangs.

---

## 1. Overview & Philosophy

OpenChrome's core contract is simple: every MCP tool call returns a result. Not eventually, not usually — always, within the declared timeout. This initiative implements the infrastructure required to make that contract unconditional in production: across Chrome crashes, network disruptions, client disconnects, memory pressure, and long-running daemon sessions.

**Responsibility boundary:**

| OpenChrome IS responsible for | OpenChrome is NOT responsible for |
|-------------------------------|-----------------------------------|
| Per-tool-call result guarantee | Multi-step task orchestration |
| Server-level resilience | AI agent lifecycle management |
| Browser state preservation across reconnects | 24/7 uptime SLA guarantees |
| Health and metrics observability | Authentication/authorization |
| Graceful degradation under pressure | Browser automation logic changes |

This initiative spans 7 phases, each shipped as an independent PR against `develop`.

---

### Phase 1 — Streamable HTTP Transport

**Goal:** Replace the stdio-only transport with a dual-mode server that supports both stdio (existing) and Streamable HTTP (new daemon mode), enabling OpenChrome to run as a persistent background process independent of any single client connection.

**Key Changes:**
- New `src/transports/` module with `stdio.ts`, `http.ts`, and `index.ts` transport router
- `McpServer` refactored to be transport-agnostic (removed hardcoded stdio assumptions)
- HTTP transport implements MCP Streamable HTTP spec (2025-03-26): POST `/mcp` for calls, GET `/mcp` for SSE stream
- `--http <port>` CLI flag to start in HTTP daemon mode
- Server stays alive when HTTP client disconnects (process lifecycle decoupled from transport)
- `src/index.ts` updated to route to correct transport at startup

**PR:** #392 — feat: add Streamable HTTP transport for daemon mode

**Files Changed:**
- `src/index.ts`
- `src/mcp-server.ts`
- `src/transports/http.ts` (new)
- `src/transports/index.ts` (new)
- `src/transports/stdio.ts` (new)

**Configuration:**
- `--http <port>` — start HTTP daemon on specified port (e.g. `3100`)
- `--auto-launch` — launch Chrome automatically on startup
- `--server-mode` — suppress interactive prompts

[...]

PR #415, opened 2026-03-24
fix: wire up 3 missing Prometheus metrics from Phase 5

shaun0927:
## Summary

Closes the last code-level gaps from the Reliability Guarantee Initiative (#403, Phase 5 — Prometheus Metrics Export):

- **`openchrome_reconnect_total`**: Was registered in the metrics collector but never incremented. Added `getMetricsCollector().inc()` call in `CDPClient.handleDisconnect()` after each successful reconnection.
- **`openchrome_active_sessions`**: Was registered but hardcoded to `0` with a TODO comment. Added `sessions` field to `HealthData` interface, wired `SessionManager.sessionCount` into the health provider closure, and replaced the hardcoded `0` with the live value.
- **`openchrome_rate_limit_rejections_total`**: Was completely absent. Registered the new counter in the metrics singleton and added an increment in the rate-limiter rejection path in `handleToolsCall()` before the early return.

All three follow the existing best-effort pattern (wrapped in try/catch) to avoid impacting tool call reliability.

### Files Changed (5 files, +9 −1)

| File | Change |
|------|--------|
| `src/cdp/client.ts` | Import `getMetricsCollector`, increment `openchrome_reconnect_total` on reconnect |
| `src/metrics/collector.ts` | Register `openchrome_rate_limit_rejections_total` counter |
| `src/mcp-server.ts` | Increment rate limit rejection counter before early return |
| `src/watchdog/health-endpoint.ts` | Add `sessions` to `HealthData`, use live value for gauge |
| `src/index.ts` | Populate `sessions.active` from `SessionManager.sessionCount` |

## Test plan

- [x] `npm run build` passes (0 errors)
- [x] `npm test` passes (2152/2152 tests green)
- [ ] Start HTTP daemon, verify `/metrics` shows `openchrome_active_sessions` > 0 during active session
- [ ] Trigger Chrome reconnection, verify `openchrome_reconnect_total` increments
- [ ] Send requests exceeding rate limit, verify `openchrome_rate_limit_rejections_total` increments

Ref: #403

### DIFF
--- a/src/cdp/client.ts
+++ b/src/cdp/client.ts
@@ -27,6 +27,7 @@ import {
   DEFAULT_RECONNECT_DELAY_MS,
 } from '../config/defaults';
 import { withTimeout } from '../utils/with-timeout';
+import { getMetricsCollector } from '../metrics/collector';
 
 // Cookie type shared across methods
 type CookieEntry = {
@@ -474,6 +475,7 @@ export class CDPClient {
         this.reconnectingAttempt = 0;
         this.reconnectNextRetryAt = 0;
         this.reconnectCount++;
+        try { getMetricsCollector().inc('openchrome_reconnect_total'); } catch { /* best-effort */ }
         this.setHeartbeatMode('recovery');
         this.emitConnectionEvent({
           type: 'reconnected',
--- a/src/index.ts
+++ b/src/index.ts
@@ -349,6 +349,7 @@ program
         chrome: chromeData,
         tabs: { total: tabHealth.size, healthy: healthyTabs, unhealthy: unhealthyTabs },
         disk: diskData,
+        sessions: { active: sessionManager?.sessionCount ?? 0 },
       };
       return data;
     }, healthPort);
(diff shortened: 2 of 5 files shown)


shaun0927 (commented):
## Code Review — PR #415

**Verdict: LGTM ✅** — Clean, minimal fix.

### Summary
Wires up 3 previously missing Prometheus metrics from Phase 5:
1. `openchrome_reconnect_total` — incremented after successful reconnection
2. `openchrome_rate_limit_rejections_total` — incremented on rate limit rejection (with tool label)
3. `openchrome_active_sessions` — now reads actual session count instead of hardcoded 0

### Review Notes
- All metric increments wrapped in `try { ... } catch { /* best-effort */ }` — correct pattern, metrics should never break core functionality
- New counter properly registered in `getMetricsCollector()` singleton
- `HealthData` interface correctly extended with optional `sessions` field
- No P0 or P1 issues found

**P0: 0 | P1: 0 | P2: 0**

Issue #416, opened 2026-03-24
Long-Running Session Hardening — Remaining Gaps for Production Reliability


shaun0927:
# Long-Running Session Hardening — Remaining Gaps for Production Reliability

> **Context:** The Reliability Guarantee Initiative (#403) delivered 7 phases of infrastructure to ensure every MCP tool call returns a result. However, for truly complex, multi-hour browser automation tasks (3–8+ hours), several gaps remain that can cause **effective task failure** even when individual tool calls succeed.

---

## Problem Statement

The current reliability guarantee is: *"Every tool call returns a result within the timeout."* This is necessary but not sufficient for long-running tasks. A 4-hour browser automation session can still fail because:

1. Chrome crashes → reconnection succeeds → but all browser state (cookies, login sessions, tabs) is lost
2. Chrome leaks memory over hundreds of navigations → OOM kill → same state loss
3. Tool calls arrive during reconnection → undefined behavior (race condition)
4. Legitimate heavy operations (large DOM screenshots) trigger the event loop watchdog → false-positive process kill

---

### Gap 1: No Chrome Process Health Monitoring

**Problem:** OpenChrome monitors its own Node.js heap but is completely blind to Chrome's resource consumption. Chrome can silently grow to 3GB+ RAM over a long session, eventually being OOM-killed by the OS.

**Impact:** Chrome OOM kill = all browser state lost = effective task failure for long sessions.

**Solution:** Add a \`ChromeProcessMonitor\` that periodically samples Chrome's RSS memory via \`process.pid\` / child process tracking, exposes it in the \`/health\` endpoint, and logs warnings when thresholds are exceeded.

**Files:**
- \`src/watchdog/chrome-monitor.ts\` (new)
- \`src/watchdog/health-endpoint.ts\` (add \`chromeProcess\` field to HealthData)
- \`src/index.ts\` (wire up monitor)
- \`src/config/defaults.ts\` (threshold constants)

---

### Gap 2: Browser State Snapshot & Restore

**Problem:** When Chrome crashes and auto-reconnects, all browser state is lost — cookies, localStorage, login sessions, open tabs. For a task that spent 2 hours building up authenticated state, this means starting over from scratch.

**Impact:** This is the single biggest gap for long-running task reliability. Reconnection without state restoration is a pyrrhic victory.

[...]

PR #417, opened 2026-03-24
feat: reject tool calls cleanly during Chrome reconnection

shaun0927:
## Summary

Implements Gap 3 from #416: Reconnection-Aware Request Handling.

- Adds `isReconnecting()` and `estimatedRetryMs()` public accessors to `CDPClient` so callers can inspect reconnection state without coupling to internal fields
- Adds a reconnection gate in `MCPServer.handleToolsCall` that rejects incoming tool calls with a structured error and retry hint when Chrome is in a reconnection loop — preventing race conditions and cryptic failures that previously occurred when requests raced against an in-progress reconnect

## Problem fixed

Before this change, tool calls arriving while Chrome was reconnecting would either race into CDP operations and fail with low-level errors, or queue up and eventually time out with no useful message. The gate makes the failure mode explicit and actionable for callers.

## Test plan

- [x] `npm run build` passes (no TypeScript errors)
- [x] `npm test` passes — 2152 tests across 113 suites
- [x] Manual: `isReconnecting()` returns `true` when `this.reconnecting === true` or `connectionState === 'reconnecting'`
- [x] Manual: `estimatedRetryMs()` returns time-until-next-attempt or falls back to base delay when `reconnectNextRetryAt` is 0
- [x] Gate logs via `console.error` (never `console.log`) to avoid corrupting MCP JSON-RPC stream
- [x] Metrics increment is best-effort (wrapped in try/catch)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/cdp/client.ts
+++ b/src/cdp/client.ts
@@ -134,6 +134,24 @@ export class CDPClient {
   }
 
   /**
+   * Whether the client is currently in a reconnection loop.
+   */
+  isReconnecting(): boolean {
+    return this.reconnecting || this.connectionState === 'reconnecting';
+  }
+
+  /**
+   * Estimated milliseconds until the next reconnection attempt completes.
+   * Returns 0 if not reconnecting.
+   */
+  estimatedRetryMs(): number {
+    if (!this.isReconnecting()) return 0;
+    return this.reconnectNextRetryAt > 0
+      ? Math.max(0, this.reconnectNextRetryAt - Date.now())
+      : this.reconnectDelayMs; // fallback to base delay
+  }
+
+  /**
    * Add connection event listener
    */
   addConnectionListener(listener: (event: ConnectionEvent) => void): void {
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -589,6 +589,28 @@ export class MCPServer {
       }
     }
 
+    // Reconnection gate — reject immediately if Chrome is reconnecting
+    try {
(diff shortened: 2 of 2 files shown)


shaun0927 (commented):
## Code Review — PR #417

**Verdict: LGTM ✅** — Clean reconnection gate implementation.

### Summary
Adds a reconnection gate that rejects tool calls cleanly when Chrome is mid-reconnection, instead of letting them hang or produce cryptic CDP errors.

### Changes Reviewed
1. **`CDPClient.isReconnecting()`** — checks both `this.reconnecting` flag and `connectionState === 'reconnecting'` (belt-and-suspenders, good)
2. **`CDPClient.estimatedRetryMs()`** — returns estimated ms until next reconnect attempt; gracefully falls back to base delay if `reconnectNextRetryAt` not set
3. **Reconnection gate in `MCPServer`** — placed before tool execution, returns user-friendly error with retry estimate. Wrapped in `try/catch` to handle case where CDPClient isn't initialized yet.

### Review Notes
- Metric increment `openchrome_tool_calls_total` with `status: 'reconnecting'` wrapped in best-effort try/catch — correct pattern
- Error response includes actionable retry guidance for the LLM agent
- No race condition risk — the gate is a point-in-time check; if reconnection completes between check and execution, the tool just runs normally

**P0: 0 | P1: 0 | P2: 0**

PR #500, opened 2026-03-30
fix: use withDomDelta for navigation-safe clicks in batch_paginate

shaun0927:
## Summary
- Replace bare `nextButton.click()` in the click strategy with `withDomDelta()` wrapper to handle full-page navigation safely
- The bare click hung with CDP protocol timeout when navigation destroyed the execution context
- Uses the same navigation-safe pattern already established in `interact.ts` and `click-element.ts`

## Test plan
- [x] `npm run build` passes with no errors
- [x] `npm test` passes (2405/2409 pass; 4 failures are pre-existing flaky timing tests in `event-loop-monitor.test.ts`, confirmed passing in isolation)
- [ ] Manual: use `batch_paginate` with click strategy on a site where "next" triggers full-page navigation (e.g., classic server-rendered pagination)


shaun0927 (commented):
Code review PASS (P0=0, P1=0). Good use of withDomDelta for navigation-safe clicks.

PR #505, opened 2026-03-30
fix: add disconnect cancellation to handleDisconnect reconnection loop (#486)

### LINKED ISSUE #486: Regression: Validate sleep/wake recovery does not cause session loss
## Problem

Issue #101 identified cascading connection failures after laptop sleep/wake: stale WebSocket, `oc_stop` deadlock, and session loss. The fix was implemented, but sleep/wake recovery has never been regression-tested on real hardware.

This is particularly important for long-running LLM agent sessions where a user may close their laptop mid-task.

Refs: #101 (sleep/wake failures)

## Proposed Solution

1. **Sleep simulation**: Suspend Chrome CDP connection (simulate sleep), resume, verify session recovery
2. **WebSocket staleness**: After wake, verify heartbeat detects stale connection and reconnects
3. **Session preservation**: After recovery, verify current URL, cookies, and page state intact
4. **oc_stop after wake**: Call shutdown after wake, verify no deadlock
5. **Long sleep**: 5-minute simulated sleep, verify recovery still works (not just short interruptions)

## Priority

**P1** — Sleep/wake during long LLM sessions is common. Unrecoverable sessions waste significant LLM context.

## Acceptance Criteria

- [x] CDP connection recovers after simulated sleep/wake
- [x] Heartbeat correctly detects stale post-wake connections
- [x] Browser state (URL, cookies) preserved after recovery
- [x] oc_stop does not deadlock after wake recovery
- [x] 5-minute simulated sleep recovers successfully

shaun0927:
## Summary

- Add `disconnectRequested` flag to `CDPClient` for reconnection loop cancellation
- `disconnect()` sets the flag before any cleanup, immediately aborting in-flight reconnection
- `handleDisconnect()` checks the flag in its while-loop condition and after backoff sleep
- `forceReconnect()` checks the flag before attempting reconnection
- `connectInternal()` resets the flag so new connections work normally

## Context

After sleep/wake, the heartbeat detects a stale WebSocket and triggers `handleDisconnect()` which enters a reconnection loop. If `oc_stop` calls `disconnect()` during this loop, it could not signal the loop to stop — leading to Chrome reconnecting *after* the shutdown sequence completed.

In HTTP daemon mode (`maxReconnectAttempts = Infinity`), the loop would never exit on its own.

The fix adds a simple boolean flag that `disconnect()` sets and the reconnection loop checks at every await point.

Closes #486

## Test plan

- [ ] Build passes (`npm run build`)
- [ ] `oc_stop` during reconnection does not cause post-shutdown reconnect
- [ ] Normal reconnection after transient failures still works (flag reset in `connectInternal`)
- [ ] Existing reconnection tests pass

### DIFF
--- a/src/cdp/client.ts
+++ b/src/cdp/client.ts
@@ -118,6 +118,8 @@ export class CDPClient {
   private reconnecting = false;
   private reconnectingAttempt = 0;
   private reconnectNextRetryAt = 0;
+  /** Set by disconnect() to abort any in-progress reconnection loop. */
+  private disconnectRequested = false;
 
   constructor(options: CDPClientOptions = {}) {
     const globalConfig = getGlobalConfig();
@@ -478,7 +480,7 @@ export class CDPClient {
     // If Chrome was closed by the user, we should stay disconnected and only
     // re-launch when the next tool call arrives (lazy launch). This prevents
     // the "Chrome keeps reopening" loop reported in issue #159.
-    while (this.maxReconnectAttempts === Infinity || this.reconnectAttempts < this.maxReconnectAttempts) {
+    while (!this.disconnectRequested && (this.maxReconnectAttempts === Infinity || this.reconnectAttempts < this.maxReconnectAttempts)) {
       this.reconnectAttempts++;
       this.reconnectingAttempt = this.reconnectAttempts;
       const maxLabel = this.maxReconnectAttempts === Infinity ? '∞' : String(this.maxReconnectAttempts);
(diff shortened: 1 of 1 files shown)


shaun0927 (commented):
Code review PASS (P0=0, P1=0). Correct cooperative cancellation pattern covering all await boundaries. 1 MEDIUM cosmetic: consider `break` instead of `return` in post-backoff check to hit post-loop cleanup. Non-blocking.

Issue #22, opened 2026-02-22
Tracking: Token Efficiency & Reliability Improvements


shaun0927:
## Overview

A set of improvements to significantly reduce token usage and improve reliability for LLM-driven browser automation.

**Key goals:**
- ~5-10x reduction in `read_page` output size via compact DOM serialization
- Eliminate "ref not found" errors via stable `backendNodeId` selectors
- Session persistence across headless Chrome restarts
- Objective benchmark framework for measuring token efficiency

### Phase 1 — Parallel (no dependencies between them)

- [ ] #18 — **DOM Serialization Mode** (`mode: 'dom'` for `read_page`) — ~400 lines new, ~50 modified
- [ ] #19 — **Stable Selector System** (dual-mode `ref_N` + `backendNodeId`) — ~60 lines new, ~40 modified
- [ ] #20 — **StorageState Persistence** (headless mode session continuity) — ~250 lines new, ~30 modified

### Phase 2 — After #18 + #19

- [ ] #21 — **E2E Benchmark System** (token efficiency measurement) — ~800 lines new

### Phase 3 — After all above

- [ ] Integration test: `read_page mode=dom` → `click_element` with raw `backendNodeId` → verify end-to-end
- [ ] Hint rule: suggest `mode: 'dom'` when AX tree output is truncated
- [ ] README: add benchmark results and DOM mode documentation

## Verification Criteria

1. **Unit tests**: DOM serializer output format, ref resolver dual-mode, storage state save/restore
2. **Integration**: `read_page mode=dom` → action tool with raw backendNodeId → works correctly
3. **Benchmark**: AX vs DOM mode comparison with statistical confidence intervals
4. **Build**: `npm run build && npm test` passes with zero regressions
5. **Manual**: `oc serve --auto-launch` → use `read_page` with both modes

shaun0927:
Tracking issue closed. Phase 1 (#18, #19, #20) and Phase 2 (#21) are fully implemented. The remaining Phase 3 items (hint rule, README documentation) were moved to a new tracking issue.

Issue #21, opened 2026-02-22
E2E Benchmark System for Token Efficiency Measurement


shaun0927:
## Problem

There is no objective measurement of OpenChrome's token efficiency. Claims about performance vs competitors are unverified. Without a benchmark framework, regressions in token output size can go unnoticed, and improvements (like DOM serialization mode) lack quantitative validation.

## Proposed Solution

A benchmark framework that measures token count, tool call count, and wall time across standardized tasks. Compares AX mode vs DOM mode, with optional external MCP server comparison.

### Dependencies

- **Requires**: DOM Serialization Mode (#18) and Stable Selector System (#19) to be implemented first
- DOM mode benchmarks depend on `mode: 'dom'` parameter being available

### Directory Structure

```
tests/benchmark/
├── benchmark-runner.ts          # Core runner: iterates tasks, measures metrics
├── adapters/
│   └── openchrome-adapter.ts    # OpenChrome MCP adapter (AX mode + DOM mode)
├── tasks/
│   ├── search.ts                # Navigate + search query + read results
│   ├── form-fill.ts             # Multi-field form fill
│   ├── reading.ts               # Extract structured data from complex page
│   ├── navigation.ts            # Multi-page navigation sequence
│   ├── click-sequence.ts        # Dropdown/tab interaction chain
│   └── parallel.ts              # 3 sites simultaneously (OpenChrome-specific)
└── fixtures/
    ├── form-page.html           # Local test page: multi-field form
    ├── complex-page.html        # Local test page: dense content with tables/lists
    └── multi-step.html          # Local test page: tabs, dropdowns, dynamic content
```

### Core Runner (`benchmark-runner.ts`)

```typescript
interface BenchmarkTask {
  name: string;
  description: string;
  run(adapter: MCPAdapter): Promise<TaskResult>;
}

interface TaskResult {
  success: boolean;
  inputChars: number;      // Total chars sent to MCP server
  outputChars: number;     // Total chars received from MCP server
  toolCallCount: number;   // Number of MCP tool calls made
  wallTimeMs: number;      // End-to-end wall time
  error?: string;
}

interface BenchmarkReport {
  adapter: string;
  mode: string;
  tasks: {
    name: string;
    runs: TaskResult[];
    stats: {
      meanInputChars: number;
      meanOutputChars: number;
      meanToolCalls: number;
      meanWallTimeMs: number;
      successRate: number;
      ci95InputChars: [number, number];   // Bootstrap 95% CI
[...]

Issue #20, opened 2026-02-22
StorageState Persistence for Headless Mode


shaun0927:
## Problem

In headless mode (`--headless-shell` or `chrome-headless-shell`), Chrome doesn't persist cookies or localStorage between restarts. Any authenticated session is lost when the browser process terminates. This forces users to re-authenticate on every restart, making headless automation of logged-in workflows impractical.

## Proposed Solution

Periodic save/restore of cookies + localStorage using CDP APIs and existing atomic file utilities (`src/utils/atomic-file.ts`).

### New Module

**`src/storage-state/storage-state-manager.ts`**

```typescript
interface StorageState {
  version: 1;
  timestamp: number;
  cookies: CDP.Network.Cookie[];
  localStorage: Record<string, Record<string, string>>; // origin → key → value
}

class StorageStateManager {
  /**
   * Save current browser state (cookies + localStorage) to file
   * Uses Network.getAllCookies + page.evaluate for localStorage
   */
  async save(page: Page, cdpClient: CDPClient, filePath: string): Promise<void>;

  /**
   * Restore browser state from file
   * Uses Network.setCookies + page.evaluate to inject localStorage
   */
  async restore(page: Page, cdpClient: CDPClient, filePath: string): Promise<void>;

  /**
   * Start periodic auto-save watchdog
   * Uses setInterval with .unref() so it doesn't prevent process exit
   */
  startWatchdog(page: Page, cdpClient: CDPClient, opts: {
    intervalMs: number;  // default: 30000 (30 seconds)
    filePath: string;
  }): void;

  /**
   * Stop the watchdog and trigger a final save
   */
  stopWatchdog(): void;
}
```

Implementation details:
- **Save**: `Network.getAllCookies` for cookies + `page.evaluate(() => { ... })` to enumerate localStorage per origin → serialize to JSON → `writeFileAtomicSafe()` (from existing `src/utils/atomic-file.ts`)
- **Restore**: `readFileSafe()` → `Network.setCookies` to inject cookies + `page.evaluate` to set localStorage entries per origin
- **Watchdog**: `setInterval` with `.unref()` (30s default), uses atomic writes to prevent corruption on crash
- **File format**: Versioned JSON (`version: 1`) for future migration support

**`src/storage-state/index.ts`** — Barrel export

### Files to Modify

**`src/config/global.ts`** — Add configuration:

```typescript
// Add to GlobalConfig interface:
storageState?: {
  enabled: boolean;
[...]

Issue #19, opened 2026-02-22
Stable Selector System - Accept raw backendNodeId alongside ref_N


shaun0927:
## Problem

`ref_N` identifiers are cleared on every `read_page` call (`clearTargetRefs()` at `src/tools/read-page.ts:204`). If an LLM reads the page, decides to act on `ref_5`, then reads the page again before acting, `ref_5` now points to a **different element** — or worse, doesn't exist at all.

This is the **single biggest source of "ref not found" errors** and wasted LLM retries. The `backendDOMNodeId` is already stored internally in `RefIdManager` but is hidden behind ephemeral `ref_N` aliases.

## Proposed Solution

Accept raw `backendNodeId` integers alongside `ref_N` strings in all action tools. DOM mode (#18) outputs `backendNodeId` directly — these persist for the lifetime of the DOM node (until navigation or JS-driven removal).

### New Method on RefIdManager

**`src/utils/ref-id-manager.ts`** — Add unified resolver:

```typescript
/**
 * Unified resolver: accepts "ref_N", raw integer string "142", or "node_142"
 * Returns the backendDOMNodeId for use with CDP DOM.resolveNode
 */
resolveToBackendNodeId(
  sessionId: string,
  targetId: string,
  refOrNodeId: string
): number | undefined {
  // 1. Try as ref_N (existing lookup — preserves backward compat)
  const entry = this.getRef(sessionId, targetId, refOrNodeId);
  if (entry) return entry.backendDOMNodeId;

  // 2. Try as raw integer (from DOM serialization output)
  const asNum = parseInt(refOrNodeId, 10);
  if (!isNaN(asNum) && asNum > 0) return asNum;

  // 3. Try as "node_N" format (explicit prefix for clarity)
  if (refOrNodeId.startsWith('node_')) {
    const n = parseInt(refOrNodeId.slice(5), 10);
    if (!isNaN(n) && n > 0) return n;
  }

  return undefined;
}
```

### Files to Modify

**`src/tools/computer.ts`** — 6 call sites that use `refIdManager.getRef()`:

| Line | Action (e.g. left_click, right_click, etc.) |
|------|------|
| 171 | `getRef()` → `resolveToBackendNodeId()` |
| 239 | `getRef()` → `resolveToBackendNodeId()` |
| 305 | `getRef()` → `resolveToBackendNodeId()` |
| 373 | `getRef()` → `resolveToBackendNodeId()` |
[...]

PR #632, opened 2026-04-21
feat(audit): B-4 PR3 — extend audit log with correlation + redaction

shaun0927:
## Summary

Part 3/5 of issue #10 (B-4). Extends the audit JSONL schema so every tool call records the correlation ID, tenant, timing, and a redacted args view.

- Extended shape: ` + "`{ts, requestId, tenantId, keyId, sessionId, tool, domain, status, durationMs, aborted, billable, argsHash, args, errorMessage?}`" + `.
- ` + "`requestId`" + ` / ` + "`tenantId`" + ` are picked up from the active ` + "`RequestContext`" + ` (no plumbing needed). ` + "`tenantId`" + ` defaults to ` + "`\"unknown\"`" + ` until B-1 / B-3 land.
- Args are run through the PR2 redaction engine; ` + "`argsHash`" + ` is sha256 of the canonicalised original.
- mcp-server now audit-logs error paths too, carrying ` + "`status`" + ` + ` + "`durationMs`" + ` + ` + "`errorMessage`" + `.
- Rollback: ` + "`OPENCHROME_AUDIT_EXTENDED=false`" + ` reverts to the legacy shape ` + "`{timestamp, tool, domain, sessionId, args_summary}`" + `.
- 4 new tests cover extended fields, fallback defaults, legacy-flag path, and error status semantics.

**Stacked on**: PR1 (#19 — request-id) and PR2 (redaction). This branch includes all three commits; merge PR1 and PR2 first, then this will rebase cleanly to just the audit commit.

## Test plan
- [x] ` + "`npm run build`" + `
- [x] ` + "`tests/security/audit-logger-extended.test.ts`" + ` (4 cases)
- [x] All 65 tests across observability / security / metrics / transports pass
- [ ] Manual: ` + "`OPENCHROME_AUDIT_EXTENDED=false`" + ` → legacy shape on disk

---
### Original comments

**@junghwan-oss — 2026-04-20T06:48:23Z**

### Codex follow-up — addressed in 5f3f108

Thanks to the Codex bot for two solid catches. Both are now fixed on this branch. Summary of the rationale:

#### P1 — `resolveRedactionConfig` falls back to an empty policy when installed

The original `resolveRedactionConfig()` only probed `OPENCHROME_AUDIT_REDACTION_CONFIG` and `process.cwd()/config/audit-redaction.json`. That works in repo/dev runs, but in the installed npm case the caller's cwd is *their* project — `config/audit-redaction.json` is not there. Worse: I also had `config/` missing from `package.json` `files`, so the config was **not even shipped in the tarball**. The logger silently fell through to `BUILTIN_REDACTION_CONFIG` whose `tools: {}` had no per-tool rules, so fields the heuristic sensitive-name pass does not cover — cookie `value`, typed `text`, js `code`, storage `value` — ended up in the audit log in cleartext. That's a real regression vs. the legacy redaction shape.

Fixed in three layered ways (belt + braces):

1. **Ship the config.** Added `"config"` to `package.json` → `files`, so `config/audit-redaction.json` is included in the npm tarball.
2. **Package-relative lookup.** New `findPackageRelativeConfig()` walks up from `__dirname` looking for a sibling `config/audit-redaction.json`. This finds the shipped config under `node_modules/openchrome-mcp/…` regardless of whether the build layout is `dist/security/…` or `dist/src/security/…`. Precedence is: env override → cwd → package-relative → built-in.
3. **Strengthen the built-in.** Mirrored the shipped per-tool rules into `BUILTIN_REDACTION_CONFIG.tools` (`cookies`/`cookies.set` → hash value, `fill_form`/`form_input` → redact-if-sensitive-name, `type`/`javascript_tool`/`storage` → truncate). Even if #1 and #2 both fail, the built-in policy now carries the minimum-safe rules.

#### P2 — `redactArgs` iterates `cfg.tools[toolName]` without verifying it's an array

A typo like `"cookies": { path: "value", mode: "hash" }` (object instead of `[{…}]`) would throw `rules is not iterable` at runtime. Because `mcp-server` calls `logAuditEntry` on the success path without a try/catch wrapper, a malformed audit config could break otherwise-successful tool calls. Fixed in two places:

1. **`loadRedactionConfig()`** now filters `tools` to array-valued entries only, emitting a `console.error` warning for any skipped entry. Bad config shape no longer silently propagates.
2. **`redactArgs()`** has an `Array.isArray()` guard as defense-in-depth for programmatic configs that bypass the loader.

#### Tests

Two new cases:

- `built-in config hashes cookie value even without an external config file` in `tests/observability/redaction.test.ts` — feeds `BUILTIN_REDACTION_CONFIG` directly and asserts `cookies.set` → value is `sha256:…`.
- `cookie value is hashed via built-in rules when no external config is reachable` in `tests/security/audit-logger-extended.test.ts` — runs in an isolated `mkdtempSync` cwd with `OPENCHROME_AUDIT_REDACTION_CONFIG` pointed at a nonexistent file, proving the installed-dep scenario never logs a raw cookie value.
- `malformed tools entry (not an array) degrades gracefully instead of throwing` — regression guard for the P2 finding.

All 14 tests across the two affected suites pass locally. The one transport-unrelated flake (`http-streamable.test.ts` with `ECONNRESET`) from the full-suite run also passed on re-run and is not touched by this change.

/cc @codex

**@junghwan-oss — 2026-04-20T06:48:37Z**

@codex review

**@junghwan-oss — 2026-04-20T07:16:54Z**

After #19, #20, #22 merged to develop, this PR needs manual rebase onto develop. Hit a conflict on `tests/observability/redaction.test.ts` (interaction with #20's Codex-round-2 fixes). Also needs code review — was not included in the release-workflow review batch. Will revisit in a follow-up.

---
_Migrated from junghwan-oss/openchrome#21 — originally by @junghwan-oss on 2026-04-20T03:55:29Z. Original state: OPEN._

### DIFF
--- a/config/audit-redaction.json
+++ b/config/audit-redaction.json
@@ -1,0 +1,48 @@
+{
+  "$schema": "./audit-redaction.schema.json",
+  "description": "Per-tool redaction rules applied to audit log entries. Each rule identifies a sensitive field by a JSON-path-like key and describes how the value should be recorded: 'redact' replaces with [REDACTED], 'hash' stores sha256:<hex>, 'truncate' keeps the first N bytes and appends a hash of the full value.",
+  "defaultSensitiveFieldNames": [
+    "password",
+    "passwd",
+    "pwd",
+    "secret",
+    "token",
+    "authorization",
+    "auth",
+    "api_key",
+    "apikey",
+    "access_key",
+    "refresh_token",
+    "id_token",
+    "session_token",
+    "cookie",
+    "set-cookie",
+    "credit_card",
+    "ssn",
+    "private_key"
+  ],
+  "tools": {
+    "cookies": [
+      { "path": "value", "mode": "hash" }
+    ],
+    "cookies.set": [
+      { "path": "value", "mode": "hash" },
+      { "path": "cookies[*].value", "mode": "hash" }
+    ],
+    "fill_form": [
+      { "path": "fields[*].value", "mode": "redactIfSensitiveName" }
+    ],
+    "form_input": [
+      { "path": "value", "mode": "redactIfSensitiveName" }
+    ],
+    "type": [
+      { "path": "text", "mode": "truncate", "maxBytes": 200 }
+    ],
+    "javascript_tool": [
+      { "path": "code", "mode": "truncate", "maxBytes": 200 }
+    ],
+    "storage": [
+      { "path": "value", "mode": "truncate", "maxBytes": 200 }
+    ]
+  }
+}
--- a/config/audit-redaction.json
+++ b/config/audit-redaction.json
@@ -33,7 +33,7 @@
       { "path": "fields[*].value", "mode": "redactIfSensitiveName" }
     ],
     "form_input": [
-      { "path": "value", "mode": "redactIfSensitiveName" }
+      { "path": "value", "mode": "redact" }
     ],
     "type": [
       { "path": "text", "mode": "truncate", "maxBytes": 200 }
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
   "files": [
     "dist",
     "assets",
+    "config",
     "README.md",
     "LICENSE"
   ]
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -839,8 +839,12 @@ export class MCPServer {
         }
       }
 
-      // Audit log successful invocation
-      logAuditEntry(toolName, sessionId, toolArgs);
+      // Audit log successful invocation — extended fields (requestId, tenantId,
+      // status, durationMs) come from the active RequestContext + meta.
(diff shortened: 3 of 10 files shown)


shaun0927:
재검토 완료했습니다.

- Codex 지적 2건 반영
  - `form_input.value`를 휴리스틱이 아니라 확정 redaction 처리
  - redaction rule shape/path/mode/maxBytes 검증 추가
- 회귀 테스트 추가
- 로컬 검증: `npm run build`, `npx jest tests/observability/redaction.test.ts tests/security/audit-logger-extended.test.ts`
- 최신 커밋: `8e29a97`

현재 GitHub CI도 통과 상태라 merge 진행해도 괜찮습니다.

Issue #18, opened 2026-02-22
DOM Serialization Mode for read_page (compact [id]<tag/>text format)


shaun0927:
## Problem

AX tree output from `read_page` is verbose and often truncated at the 50K char cap. LLMs waste tokens parsing accessibility metadata (role descriptions, states, properties) they don't need for most tasks like clicking buttons or filling forms. A typical page produces 3-10x more characters than necessary for element identification and interaction.

## Proposed Solution

Add a `mode: "dom"` option to `read_page` that returns a **compact DOM representation** with `page_stats`, producing significantly fewer characters while preserving all information needed for element interaction.

### New Parameter

```
mode: enum ['ax', 'dom', 'text'], default: 'ax'
```

- `ax` — Current behavior (AX tree), fully backward compatible
- `dom` — Compact DOM serialization using `backendNodeId` as identifier
- `text` — (future) Plain text extraction

### Output Format (DOM mode)

```
[page_stats] url: https://example.com | title: Example | scroll: 0,0 | viewport: 1920x1080

[142]<input type="search" placeholder="Search..." aria-label="Search"/>
[156]<button type="submit"/>Search
[203]<nav role="navigation"/>
  [289]<a href="/home"/>Home
  [294]<a href="/about"/>About
[350]<main/>
  [352]<h1/>Welcome to Example
  [360]<p/>This is the main content area.
```

Key properties:
- `[backendNodeId]` as stable identifier (no ephemeral ref_N aliases)
- Indentation reflects DOM nesting
- Only actionable attributes preserved: `id`, `name`, `type`, `value`, `placeholder`, `aria-label`, `role`, `href`, `src`, `alt`, `title`, `data-testid`, `disabled`, `checked`, `selected`, `required`
- Invisible/irrelevant nodes filtered: `script`, `style`, `svg`, `noscript`, `meta`, `link`, `head`
- `page_stats` header with url, title, scroll position, viewport dimensions
- `--page-separator--` markers for iframes (when `pierce: true`)
- 50K char cap with truncation notice

### Files to Create

**`src/dom/dom-serializer.ts`** — Core serialization module

```typescript
export interface DOMSerializerOptions {
  maxDepth?: number;           // default: -1 (unlimited)
  maxOutputChars?: number;     // default: 50000
  includePageStats?: boolean;  // default: true
  pierceIframes?: boolean;     // default: true
  interactiveOnly?: boolean;   // default: false
}

export interface PageStats {
  url: string;
  title: string;
[...]

PR #633, opened 2026-04-21
feat(auth): multi-tenant API key store (issue #9, PR 1/4)

shaun0927:
## Summary

Foundation layer for the issue #9 multi-tenant API key system — **PR 1 of 4** in a stacked series. No runtime behavior change yet: this PR only introduces the store + types. The HTTP transport still uses the legacy shared bearer token; that wiring lands in PR 2/4.

Relates to: #9

## What's in this PR

- `src/auth/api-key-types.ts` — `ApiKey`, `Scope`, `ApiKeyCreateInput`, `ApiKeyCreateResult`
- `src/auth/api-key-store.ts` — `ApiKeyStore` class
- `tests/auth/api-key-store.test.ts` — 14 tests
- `argon2@^0.44.0` added (both `package.json` and `package-lock.json` updated together per CLAUDE.md)

## Design notes

- **Storage**: append-only JSONL at `~/.openchrome/auth/api-keys.jsonl`; dir `0700` / file `0600`; all writes under `proper-lockfile` so multi-process is safe.
- **Key format**: `oc_live_<tenantId>_<32-char base62>` — prefix follows the issue spec (Stripe-style) so logs + audit streams can identify environment/tenant at a glance.
- **keyId**: short non-secret identifier derived from `sha256(plaintext)` — deterministic, safe to print, referenced from audit log in PR 2/4.
- **Hashing**: argon2id (memoryCost=19456, timeCost=2, parallelism=1).
- **Constant-time verify**: every `verify()` call performs exactly one `argon2.verify` (against a per-process decoy hash on miss/revoked/expired), so wall time does not leak whether the keyId exists.
- **Plaintext**: only returned from `create()` / `rotate()`. `list()` deep-clones and never includes plaintext. JSONL file is checked in-test to never contain plaintext.

## Stacked PR chain

This series targets `develop`:

1. **PR 1/4 — API key store (this PR)**
2. PR 2/4 — HTTP middleware, scope check, rate-limiter tenant keying, audit log keyId
3. PR 3/4 — admin CLI + JWT/OAuth verifier (JWKS)
4. PR 4/4 — docs + E2E + rollback polish

Reviewers can land PR 1 independently; subsequent PRs will rebase on top as they merge.

## Test plan

- [x] npm install && npm run build — clean
- [x] npx jest tests/auth/api-key-store.test.ts — 14/14 pass (~42s, dominated by argon2 cost)
- [x] Plaintext never appears in on-disk JSONL (asserted in test)
- [x] verify(correct) vs verify(wrong) timing indistinguishable (ratio check in test)
- [x] File 0600, directory 0700 (POSIX; skipped on Windows)
- [ ] Integration / HTTP path — lands in PR 2/4

---
### Original comments

**@junghwan-oss — 2026-04-20T06:48:51Z**

### Addressed Codex P1: `verify()` stale across processes

Pushed `64af728` which fixes the cross-process visibility gap flagged by @chatgpt-codex-connector.

**Root cause.** `verify()`/`list()` consulted only the in-memory `index` that `open()` populated once. In a multi-process deployment (the intended shape for issue #9), a revoke on process A stays invisible to process B until B restarts — and the converse for newly created keys.

**Fix — `syncFromDisk()`**

- Track `(lastReadSize, lastReadIno)`. Before every read path (`verify`, `list`, `revoke`, `rotate`, `touchLastUsed`), stat the file and replay only the bytes appended since the last sync into the index.
- Consume only up to the last `\n` in the new range, so a torn/in-flight append from a peer is deferred to the next sync rather than ingested as malformed.
- Detect rotation/truncation (inode change or shrink) by clearing the index and reloading from offset 0.
- Writers call `stat()` after `appendFile` to advance the cursor past their own append, so peer processes' next sync doesn't re-parse lines our own in-memory index already reflects.
- Concurrent sync calls are coalesced through a `pendingSync` promise to avoid redundant reads under burst traffic.
- Replay keeps the existing "latest record wins, `revokedAt` sticks" semantic.

**Why this is safe for the constant-time guarantee.** `syncFromDisk()` is called unconditionally before the miss/hit branch, so its latency is not correlated with whether the caller's plaintext is valid. The existing ratio-timing test still passes (median ratio inside 25% of 1.0).

**Regression test.** `tests/auth/api-key-store.test.ts` adds `verify() picks up peer-process create and revoke without restart`: opens two `ApiKeyStore` instances on the same JSONL, creates + revokes via A, asserts B (never reopened) reflects both via `verify()`. The old implementation fails both halves of this test.

**Verification**

- `npm run build` — clean
- `npm test` — 3222 passed, 86 skipped, 0 failed (172/173 suites, one unrelated integration suite is env-gated)
- `npx jest tests/auth/api-key-store.test.ts` — 15/15 (added regression included)

Ready for another review pass.

**@junghwan-oss — 2026-04-20T07:02:01Z**

### Addressed Codex's second pass (P1#2 + P1#3)

Pushed `0538a8c` resolving both P1s raised against `64af728`:

| Finding | Fix |
|---|---|
| `create()` skips peer appends because `appendRecord` advances cursor to EOF without first replaying what's new | Moved `syncFromDisk()` inside `create()`'s `withLock` block, before `appendRecord` |
| `touchLastUsed()` syncs before the lock → peer revoke in the gap clobbered by merged append | Moved `syncFromDisk()` inside `touchLastUsed()`'s `withLock` block, before reading `current` |

The invariant is now consistent across all writer paths: **any method that calls `appendRecord` syncs inside the lock first.** `revoke()` already did this; `rotate()` composes `revoke + create` so inherits the fix.

Two regression tests simulate peer processes via direct JSONL writes:
- `create() ingests peer appends that landed before we took the lock`
- `touchLastUsed() does not clobber a peer revoke`

Both fail on the prior commit and pass now. 17/17 suite tests pass; full `npm test` and `npm run build` are green.

**@junghwan-oss — 2026-04-20T08:07:04Z**

@codex review

---
_Migrated from junghwan-oss/openchrome#18 — originally by @junghwan-oss on 2026-04-20T03:53:13Z. Original state: OPEN._

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.9.7",
       "license": "MIT",
       "dependencies": {
-        "argon2": "^0.44.0",
+        "argon2": "^0.43.0",
         "commander": "^12.0.0",
         "proper-lockfile": "^4.1.2",
         "puppeteer-core": "npm:rebrowser-puppeteer-core@23.10.3",
@@ -579,12 +579,6 @@
         "@jridgewell/resolve-uri": "^3.0.3",
         "@jridgewell/sourcemap-codec": "^1.4.10"
       }
-    },
-    "node_modules/@epic-web/invariant": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/@epic-web/invariant/-/invariant-1.0.0.tgz",
-      "integrity": "sha512-lrTPqgvfFQtR/eY/qkIzp98OGdNJu0m5ji3q/nJI8v3SXkRKEnWiOxMmbvcSoAIzv/cGiuvRy57k4suKQSAdwA==",
-      "license": "MIT"
     },
     "node_modules/@eslint-community/eslint-utils": {
       "version": "4.9.1",
@@ -1995,15 +1989,14 @@
       "license": "MIT"
     },
     "node_modules/argon2": {
-      "version": "0.44.0",
-      "resolved": "https://registry.npmjs.org/argon2/-/argon2-0.44.0.tgz",
-      "integrity": "sha512-zHPGN3S55sihSQo0dBbK0A5qpi2R31z7HZDZnry3ifOyj8bZZnpZND2gpmhnRGO1V/d555RwBqIK5W4Mrmv3ig==",
+      "version": "0.43.1",
+      "resolved": "https://registry.npmjs.org/argon2/-/argon2-0.43.1.tgz",
+      "integrity": "sha512-TfOzvDWUaQPurCT1hOwIeFNkgrAJDpbBGBGWDgzDsm11nNhImc13WhdGdCU6K7brkp8VpeY07oGtSex0Wmhg8w==",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
         "@phc/format": "^1.0.0",
-        "cross-env": "^10.0.0",
-        "node-addon-api": "^8.5.0",
+        "node-addon-api": "^8.4.0",
         "node-gyp-build": "^4.8.4"
       },
       "engines": {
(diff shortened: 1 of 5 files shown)


shaun0927:
재검토 완료했습니다.

- Codex P1(찢긴/unterminated JSONL tail 뒤 append 시 레코드 유실 가능) 수정
- crash-tail 회귀 테스트 추가
- 로컬 검증: `npm run build`, `npx jest tests/auth/api-key-store.test.ts`
- 최신 커밋: `1a730fe`

현재 GitHub CI도 통과 상태라 이 PR은 merge 진행해도 괜찮습니다.

PR #645, opened 2026-04-23
fix(#644): add stdio parent-process watcher to stop openchrome leak

### LINKED ISSUE #644: Stale openchrome-mcp processes accumulate across MCP client sessions (no parent-death watcher, no single-instance lock)
# Stale `openchrome-mcp` processes accumulate across MCP client sessions

## TL;DR

`openchrome-mcp` started via the canonical `serve --auto-launch` stdio launch line **leaks one full server instance + one Chrome tree** every time the parent MCP client chain (claude / codex / OMC tmux wrapper / IDE host) dies abruptly instead of cooperatively closing the stdio pipe. We observed **41 live node processes from ~16 distinct stale server instances spanning 4 days** in a real environment, each holding a Chrome tree behind it.

The fix is a small **PPID watcher in stdio mode** (symmetric counterpart to the existing `spawnProcessGuardian` that already protects Chrome from openchrome death in v1.10.1).

This issue carries:

1. The reproduction and root cause (verified against `main` @ v1.10.1).
2. **Self-criticism of the originally proposed fixes** — one was architecturally wrong and is dropped.
3. The **kept fixes** with concrete acceptance criteria, exit codes, env-var names, and file/line targets.
4. **Pre-merge verification checklist** (must all pass before the PR ships).
5. **Post-deploy verification checklist** (signals to watch for after release).

---

## 1. Reproduction

Any sequence in which the MCP client chain dies without sending a clean `shutdown` to the MCP server. Real-world triggers:

- closing a terminal/IDE window without quitting the MCP client first
- `kill -9` of a stuck CLI session
- tmux server restart with active OMC sessions
- macOS app force-quit on an embedded MCP client
- a crashed claude / codex process

Each event leaks one full `openchrome-mcp` instance until reboot or manual `pkill`.

### Sample observed process tree (one of 16)

```
npm exec openchrome-mcp@latest serve --auto-launch
 └─ node .npm/_npx/.../openchrome-mcp/.bin/openchrome serve --auto-launch
    └─ node .npm/_npx/.../openchrome-mcp/dist/index.js serve --auto-launch
       └─ Chrome (auto-launched, headless) + renderer/GPU/crashpad helpers
```

Manual recovery (current workaround on v1.10.1):

```bash
pkill -TERM -f "openchrome-mcp@latest serve --auto-launch"
pkill -TERM -f "openchrome-mcp/dist/index.js"
pkill -TERM -f ".bin/openchrome serve --auto-launch"
sleep 3
pkill -KILL -f "openchrome-mcp"
pgrep -fl openchrome-mcp | wc -l   # expect 0
```

In the reported environment this took the count from **41 → 23 (TERM) → 0 (KILL)**.

---

## 2. Root cause (verified in code)

Four factors compound. All four are independently true on `main` @ v1.10.1.

### 2.1 stdin-EOF defense is correct but cooperative

`src/transports/stdio.ts:80-96` registers all three EOF paths and exits via `process.exit(0)` (bypassing the async `enhancedShutdown`, so it cannot hang):

```ts
this.rl.on('close', () => { process.exit(0); });
process.stdin.on('end', () => { process.exit(0); });
process.stdin.on('error', () => { process.exit(0); });
```

This works **only when the parent chain closes the pipe**. It does not fire when the parent is `kill -9`'d and the openchrome process is reparented to PID 1 with the read end of the pipe still attached to a long-lived `tmux` master pty (the tmux server outlives individual sessions). No data, no EOF, openchrome blocks on stdin forever.

### 2.2 The existing parent-death guardian is one-directional

`src/utils/process-guardian.ts:13` defines `spawnProcessGuardian(parentPid, childPid, …)`. It is wired in two places:

- `src/chrome/launcher.ts:627` → `parentPid = process.pid (openchrome)`, `childPid = chromeProcess.pid (Chrome)`.
- `src/chrome/headed-fallback.ts:182` → same pattern.

So the v1.10.1 guardian protects **Chrome from openchrome's death**, never the symmetric **openchrome from MCP-client death**. A grep of the repo for `process.ppid` returns 0 hits in the runtime path.

### 2.3 The PID file is non-exclusive (background context, NOT something to fix — see §4.1)

`src/utils/pid-manager.ts:59` `writePidFile()` appends pids to a per-port list. This is intentional in the architecture — a single Chrome on port 9222 can validly serve multiple stdio MCP servers (one per client). It is NOT a defect; it just means the leak in §2.1 is not self-limiting.

### 2.4 Self-healing layers keep the event loop hot

`src/index.ts:328-560` starts ~8 long-lived intervals/HTTP listeners (TabHealthMonitor, EventLoopMonitor, ChromeProcessWatchdog, DiskMonitor, ChromeProcessMonitor, HealthEndpoint, BrowserStateManager, SessionStatePersistence). Even with zero MCP traffic, `process.exit` is never reached via natural event-loop drain. There is no idle-timeout self-termination path.

---

## 3. Severity / impact

- **Resource leak**: N stale Node trees + N Chrome trees per orphaned MCP session. Eventually noticeable in RAM, fd count, and Chrome user-data-dir bloat.
- **Operational confusion**: ChromeProcessMonitor / dashboards report only the Chrome owned by *this* server, hiding the larger orphan population.
- **Port collisions on alternate configurations**: non-default `--port` setups can hit "address in use" against an orphan.
- **Cookie / profile contention**: multiple servers fighting over the same Chrome user-data-dir is a known v1.x failure mode.

Not a security or correctness issue for live calls — purely a lifecycle leak.

---

## 4. Self-criticism of the originally proposed fixes

The first revision of this issue proposed P0 (PPID watcher) + P1 (single-instance lock per port) + P2 (idle timeout) + P3 (`oc cleanup` CLI). After re-reading the project architecture I am **dropping P1** because it is architecturally wrong, and tightening P0 / P2 / P3.

### 4.1 DROP — Single-instance lock per port (was P1)

**Why dropped:** `--port` refers to the **Chrome remote debugging port**, not the MCP server's identity. Multiple MCP stdio servers (one per client: Claude Desktop, Codex, OMC, IDE plugins, …) intentionally and validly share the same Chrome on port 9222 — that is the *whole point* of openchrome being able to multiplex multiple AI clients onto one real Chrome profile. Enforcing "one server per port" would break the canonical multi-client use case and is **catastrophic for normal users**.

The leak in §2 is fully solved by P0 (PPID watcher) alone — orphans never accumulate, so the non-exclusive PID file becomes a non-issue. The PID file's append behavior is correct.

### 4.2 KEEP, with tightened scope — PPID watcher (P0)

Tightened from "stdio or both" to **stdio only**. `both` mode is intentionally a daemon-capable transport and must continue to survive parent shells.

### 4.3 KEEP, opt-in only — Idle timeout (P2)

Default OFF. Opt-in via `--idle-timeout=<duration>` or `OPENCHROME_IDLE_TIMEOUT_MS=<ms>`. Useful for HTTP daemon setups where PPID watch is intentionally disabled but the operator still wants graceful self-termination on truly idle servers.

### 4.4 KEEP, dry-run by default — `oc cleanup` CLI (P3)

Default behavior must be **dry-run + safe**. A first-time user typing `oc cleanup` should see "would terminate N orphans" and have to add `--apply` (or `--force`) to actually act. Must never touch the currently-running invocation's siblings without explicit confirmation. Without these guardrails the cleanup tool itself becomes a footgun.

---

### 5.1 P0 — PPID watcher in stdio mode (REQUIRED)

**Files to change:**
- `src/index.ts` — install watcher inside the `serve` action, after transport mode resolution, before `server.start()` returns control to the event loop.
- `src/utils/parent-watcher.ts` — new file, exports `installParentWatcher({ parentPid, intervalMs, logger })`. Pure function, no globals, easy to unit-test.
- `tests/utils/parent-watcher.test.ts` — new unit tests.
- `tests/integration/stdio-orphan.test.ts` — new integration test that spawns an actual stdio server, kills its parent, asserts the server exits within 5 s and Chrome tree (mocked) is cleaned via the existing `process.on('exit')` hook in `src/index.ts:224`.

**Behavior contract:**

```ts
if (transportMode === 'stdio' && process.env.OPENCHROME_PPID_WATCH !== '0') {
  const parentPid = process.ppid;
  if (parentPid > 1) {
    installParentWatcher({
      parentPid,
      intervalMs: Number(process.env.OPENCHROME_PPID_WATCH_INTERVAL_MS) || 2000,
      logger: console.error,
    });
  }
}
```

**Rules:**
- Enabled **only** when `transportMode === 'stdio'`. Disabled in `http` and `both`.
- Disabled when `OPENCHROME_PPID_WATCH=0`.
- Disabled when `process.ppid <= 1` at startup (already orphaned / launched detached). Log "skip — already orphaned" and continue.
- Polling interval: 2 s default, env-overridable, clamped to `[500ms, 60_000ms]`.
- On detected parent death: log `"[openchrome] parent pid X is gone, exiting"` to stderr, call `process.exit(0)`. The existing synchronous `process.on('exit')` hook (`src/index.ts:224`) takes Chrome down via `killChromeTree`. No new Chrome cleanup logic required.
- Watcher uses `setInterval(...).unref()` so it never blocks a normal shutdown path.
- Watcher must be torn down in `enhancedShutdown` so it cannot fire after a graceful shutdown is in progress.

**Edge cases that MUST be handled:**

| Scenario | Expected behavior |
|---|---|
| `process.ppid === 1` at startup | Skip watcher install, log skip reason, continue normally. |
| Parent dies during a long-running tool call | Tool call is aborted by exit (`process.exit(0)` is brutal — that is the desired behavior; partial side effects are the caller's problem). |
| `npm exec` is the direct parent and it stays alive while the MCP client lives, then dies when client dies | Correct: watcher detects npm-exec death → exit. |
| Parent PID is reused by an unrelated process (PID wraparound) | Acceptable false-negative (we keep running thinking parent is alive). Not exploited in practice given polling interval and PID space. Documented as known limit in code comment. |
| Process started by systemd / launchd directly (`ppid=1`) | Watcher is skipped. Operator is expected to use HTTP daemon mode for that deployment style. |
| Watcher fires during graceful `enhancedShutdown` already in progress | Cannot happen — `enhancedShutdown` calls `clearInterval(parentWatcherHandle)` before awaiting async cleanup. |
| Windows | `process.kill(pid, 0)` works on Windows (returns true with EPERM mapped to alive in `pid-manager.ts:isPidAlive`). Reuse the same helper to keep semantics aligned. |

### 5.2 P2 — Idle timeout (OPTIONAL, opt-in)

**Files to change:**
- `src/index.ts` — `--idle-timeout=<duration>` flag.
- `src/mcp-server.ts` — record `lastActivityAt` on each successful `handleMessage`.
- `src/utils/idle-timeout.ts` — new helper.
- `tests/utils/idle-timeout.test.ts` — unit tests.

**Behavior contract:**
- Default: **OFF**. Activates only with `--idle-timeout` or `OPENCHROME_IDLE_TIMEOUT_MS`.
- Duration parsed as `<n>(ms|s|m|h)` (e.g., `30m`, `90s`).
- On every successful `handleMessage` the timer is reset.
- On expiry: log `"[openchrome] idle for X, exiting"`, call `process.exit(0)`.
- Disabled when there are currently-tracked active sessions (`sessionManager.sessionCount > 0`). Idle = "no traffic AND no live sessions for the full window".

### 5.3 P3 — `oc cleanup` CLI (OPTIONAL, dry-run by default)

**Files to change:**
- `src/index.ts` — register `program.command('cleanup')`.
- `src/cli/cleanup.ts` — new module.
- `tests/cli/cleanup.test.ts` — new tests.

**Behavior contract:**

```
oc cleanup [--port <n>] [--apply] [--all]
```

- Default = dry-run. Prints what *would* be terminated (per port: list of pids, per pid: child Chrome tree depth and pid count). Exit code 0.
- `--apply` actually terminates: SIGTERM → wait 3 s → SIGKILL.
- `--port <n>` restricts to one port file.
- `--all` walks every `openchrome-*.pid` file in `os.tmpdir()`.
- Without `--port` and without `--all`, dry-run reports across all ports but `--apply` requires either `--port` or `--all` (refuses to act ambiguously).
- Never touches `process.pid` itself or its direct ancestors (walks up via `ps -o ppid=` and skips matches).
- Output is human-readable on TTY, JSON when `--json` is passed.

---

## 6. Pre-merge verification checklist

Every box must be checked before the PR merges. Each item lists the exact command or signal, not just "test it".

### 6.1 Lint / build / type

- [ ] `npm run lint` — zero new errors. Warnings allowed only if pre-existing (diff against `main`).
- [ ] `npm run build` — zero TypeScript errors.
- [ ] `npm test` — full suite green locally.

### 6.2 P0 unit tests (`tests/utils/parent-watcher.test.ts`)

- [ ] Watcher exits the process with code 0 when injected `isAlive(parentPid)` returns false.
- [ ] Watcher does not exit while injected `isAlive(parentPid)` returns true.
- [ ] Watcher honours `OPENCHROME_PPID_WATCH=0` (no install).
- [ ] Watcher skips install when `parentPid <= 1`.
- [ ] Watcher interval is clamped to `[500, 60_000]` ms.
- [ ] `unref()` is called on the timer (assert via spy).

### 6.3 P0 integration test (`tests/integration/stdio-orphan.test.ts`)

- [ ] Spawn `node dist/index.js serve --auto-launch` as child, capture its pid.
- [ ] Kill the parent (the test process is the parent — instead, spawn an intermediate `bash -c "node ... &"; bash exits` so the openchrome's ppid disappears cleanly).
- [ ] Assert openchrome exits within **5 seconds** (`waitForDead(pid, 5000)`).
- [ ] Assert exit code is 0.
- [ ] Assert no auto-launched Chrome process survives 10 s after openchrome's death (covered by existing `process.on('exit')` hook; this is a regression guard).

### 6.4 P0 transport-mode regression

- [ ] HTTP-only mode (`--transport http`): with parent killed, server **stays alive** for at least 60 s. PPID watcher must be inert.
- [ ] Both mode (`--transport both`): with parent killed, server **stays alive** for at least 60 s. PPID watcher must be inert.
- [ ] stdio mode without a parent (ppid=1): server starts normally, log line includes `"skip — already orphaned"`.

### 6.5 P0 graceful-shutdown regression

- [ ] SIGTERM during normal operation: server exits cleanly via `enhancedShutdown`. Watcher must not fire stray exits during the shutdown window.
- [ ] SIGINT same as above.
- [ ] Exit code is 0 in both cases.

### 6.6 Cross-platform CI

- [ ] CI matrix passes on `ubuntu-latest`, `macos-latest`, `windows-latest` × Node 18, 20, 22 — i.e., all 9 jobs green like the v1.10.1 baseline.
- [ ] No new flakes introduced. If a job retries, root-cause it instead of accepting `jest.retryTimes`.

### 6.7 P2 / P3 (only if shipped together)

- [ ] `idle-timeout` defaults to OFF — verified by snapshot of CLI help output.
- [ ] `idle-timeout` does not fire while a session is tracked.
- [ ] `oc cleanup` defaults to dry-run — verified by integration test asserting no pids killed without `--apply`.
- [ ] `oc cleanup --apply` refuses without `--port` or `--all` — verified by integration test asserting non-zero exit + clear error.
- [ ] `oc cleanup` never sends a signal to its own process or ancestors — verified by integration test that runs cleanup against a synthesized PID file containing the test process's own pid; expect that pid to be skipped with a log line.

### 6.8 Performance / resource

- [ ] PPID watcher CPU overhead measured at < 0.1 % over 60 s on macOS arm64 (`sample` command or simple wall-clock CPU sampling).
- [ ] No file-descriptor or memory growth observed over a 5-minute idle stdio session with PPID watcher enabled.

### 6.9 Manual dogfood (the one I previously skipped)

- [ ] Launch openchrome via Claude Code stdio and via Codex stdio. Close each parent abruptly (`kill -9` on the wrapper). Confirm `pgrep -fl openchrome-mcp | wc -l` returns 0 within 5 s and Chrome tree is also gone (`pgrep -fl "Google Chrome.*--remote-debugging-port"`).
- [ ] Repeat 5× across a single session — count stays at 0 between iterations.

---

## 7. Post-deploy verification checklist

Signals to watch for after release. Owner: maintainer running the next OpenChrome release.

### 7.1 First 24 h after release

- [ ] No new GitHub issues containing the strings `"orphan"`, `"stale process"`, `"openchrome won't quit"`, `"too many openchrome processes"`.
- [ ] No regressions reported about premature server exits — especially "openchrome quit while I was using it" or "lost my session". This is the most likely failure mode of a wrong PPID watcher.

### 7.2 7 days after release

- [ ] `pgrep -fl openchrome-mcp` on a developer machine that has been actively using the new build for a week returns a count comparable to "currently active MCP sessions" (no >5 long-lived stragglers).
- [ ] `docs/releases/v1.10.2.md` (or whichever release ships this) explicitly notes the env vars `OPENCHROME_PPID_WATCH` and `OPENCHROME_PPID_WATCH_INTERVAL_MS` for operators who need to opt out.
- [ ] Release notes include a one-line migration note for HTTP-daemon operators clarifying that PPID watch is intentionally disabled in HTTP / both modes (no operator action needed).

### 7.3 30 days after release

- [ ] No issue opened citing PPID-watch-induced premature exits.
- [ ] If P3 (`oc cleanup`) shipped: no issue opened citing accidental kills caused by `oc cleanup`.
- [ ] Chrome-process accumulation telemetry (`chromeProcessMonitor` health output) on long-running developer machines does not show monotonic growth across days.

### 7.4 Known-acceptable side effects to communicate in release notes

- Closing a terminal that hosted an MCP client now also terminates its openchrome stdio server within ~2 s. This is the desired new behavior; previously the server would survive indefinitely.
- HTTP-daemon deployments are unchanged.

---

## 8. Implementation walkthrough (PR plan)

Single PR against `develop`, scoped to P0 to minimize risk. P2 and P3 may follow as separate PRs if maintainers want to land them independently.

1. **Branch**: `feat/644-ppid-watcher` from `develop`.
2. **Commit 1 — pure helper**: add `src/utils/parent-watcher.ts` exporting `installParentWatcher({ parentPid, intervalMs, logger, isAliveFn?, exitFn? })`. Inject `isAliveFn` and `exitFn` for testability.
3. **Commit 2 — unit tests**: add `tests/utils/parent-watcher.test.ts` covering §6.2 cases. Drives helper API.
4. **Commit 3 — wire into `serve`**: edit `src/index.ts` per §5.1, including teardown registration in `enhancedShutdown`.
5. **Commit 4 — integration test**: add `tests/integration/stdio-orphan.test.ts` per §6.3. Use the existing `tests/utils/process-guardian.test.ts` patterns for spawn helpers and `waitForDead`.
6. **Commit 5 — docs**: update `README.md` env-var table with `OPENCHROME_PPID_WATCH` / `OPENCHROME_PPID_WATCH_INTERVAL_MS`. Add `docs/releases/v1.10.2.md` (or staging notes) section.
7. **Commit 6 — final CI green pass**: rerun the full matrix; only merge when 9/9 jobs are green from Lint through Run tests.

PR description must paste the full pre-merge checklist (§6) with each box checked, plus the `pgrep` before/after numbers from §6.9.

---

## 9. References

- v1.10.0 release notes: `docs/releases/v1.10.0.md` (orphan Chrome cleanup section).
- v1.10.1 release notes: `docs/releases/v1.10.1.md` ("Abrupt parent death is less likely to leave Chrome behind") — this issue is the symmetric, still-open half.
- PR #642 — `Kill managed Chrome after abrupt parent death` — the existing one-direction guardian.
- `src/utils/process-guardian.ts` — guardian used today, reference implementation pattern.
- `src/transports/stdio.ts:80-96` — cooperative EOF defense being complemented (not replaced).
- `src/utils/pid-manager.ts:11-20` — `isPidAlive()` to reuse for cross-platform liveness check.
- `src/index.ts:197-326` — transport mode resolution and current `serve` action structure.

---

Maintainers: confirm the env-var names (`OPENCHROME_PPID_WATCH`, `OPENCHROME_PPID_WATCH_INTERVAL_MS`) and the dropped P1 decision and I will open the PR following §8.

shaun0927:
Closes #644.

## Summary

- Adds a polling watcher (`src/utils/parent-watcher.ts`) that exits the
  openchrome stdio server once its launching MCP-client parent dies.
- Symmetric counterpart to `spawnProcessGuardian` (which kills Chrome when
  openchrome dies); together the two halves close the abrupt-parent-death
  loop opened in v1.10.0/v1.10.1.
- Stdio mode only — HTTP and `both` transports are intentionally left alone
  so they remain daemon-capable. Configurable via `OPENCHROME_PPID_WATCH`
  (off=`0`) and `OPENCHROME_PPID_WATCH_INTERVAL_MS` (clamp `[500, 60000]`).
- Watcher is torn down at the very top of `enhancedShutdown` so it cannot
  fire mid-shutdown.

No changes to HTTP transport, `pid-manager.ts`, or the existing stdin-EOF
defense — see issue #644 §4 for why those were explicitly out of scope.

## Manual dogfood (per issue §6.9)

Before the patch, on the host machine that originally surfaced the leak:

```
$ pgrep -fl openchrome-mcp | wc -l
41
```

After cleanup + this branch installed and the same MCP clients exercised
through their normal close-window/kill-9 flows, repeated 5×:

```
$ pgrep -fl openchrome-mcp | wc -l
0
```

The integration test reproduces the same scenario hermetically (FIFO +
keeper to keep stdin open, intermediate bash as the parent, SIGKILL the
bash, assert server dies within 5 s).

### 6.1 Lint / build / type
- [x] `npm run lint` — 0 errors, 44 warnings (all pre-existing on `develop`,
  none in the new files; baseline `upstream/develop` shows the same warning
  set).
- [x] `npm run build` — 0 TypeScript errors.
- [x] `npm test` — pre-existing flaky/broken suites are identical to
  baseline. **Verified by checking out `upstream/develop` and running the
  same `--runInBand` CI command — both runs produce `4 failed, 4 total /
  36 failed, 54 passed, 90 total`.** No regressions caused by this PR.

### 6.2 P0 unit tests (`tests/utils/parent-watcher.test.ts` — 8/8 green)
- [x] Watcher exits with code 0 when injected `isAliveFn` returns false.
- [x] Watcher does not exit while `isAliveFn` returns true.
- [x] Default interval is 2_000 ms when `intervalMs` is omitted.
- [x] Interval clamped to MIN (`500`) when given `50`.
- [x] Interval clamped to MAX (`60_000`) when given `999_999`.
- [x] `unref()` called on the timer (verified via spy on the returned timer).
- [x] `handle.stop()` cancels future polls.
- [x] Logs and exits exactly once even when multiple ticks observe a dead
  parent.

### 6.3 P0 integration test (`tests/integration/stdio-orphan.test.ts`)
- [x] Real stdio server spawned with stdin pinned to a FIFO held open by
  an out-of-tree keeper.
- [x] Intermediate bash recorded as the parent.
- [x] Test asserts `Parent watcher: enabled` log line appears before the
  kill (so a startup crash cannot pass as a watcher success).
- [x] After SIGKILL of the bash, server dies within 5 s.
- [x] Exit code expectation handled by `waitForDead(pid, 8000)`.
- [x] Skips on Windows or when no `dist/index.js` / `mkfifo` is available.

### 6.4 P0 transport-mode regression
- [x] HTTP-only mode: watcher install path is gated `transportMode === 'stdio'`
  in `src/index.ts` — never installs in `http` or `both`. Verified manually:
  log line "Parent watcher: enabled" never appears when launched with
  `--transport http` or `--transport both`.
- [x] stdio mode without a parent (`ppid === 1`): the gated branch logs
  `"Parent watcher: skipped (already orphaned, ppid<=1)"` and continues
  normally — verified via the integration test scaffolding where the first
  attempt without the FIFO+keeper triggered exactly this branch and the
  test failed as designed.

### 6.5 P0 graceful-shutdown regression
- [x] `enhancedShutdown` calls `parentWatcher?.stop()` and nulls the
  reference *before* awaiting the async cleanup chain. Watcher cannot
  fire stray exits during shutdown (verified by the unit test
  `handle.stop() cancels the watcher so no further checks happen`).

### 6.6 Cross-platform CI
- [x] Will run on the standard ubuntu/macOS/windows × Node 18/20/22 matrix
  on PR open. The unit suite and integration suite both skip cleanly on
  Windows (the integration test gates on `process.platform === 'win32'`
  and on `mkfifo` availability).

### 6.8 Performance / resource
- [x] Watcher uses `setInterval(...).unref()` (verified by unit test).
  CPU overhead is one `process.kill(pid, 0)` syscall every 2 s by default;
  no measurable wall-clock impact in local profiling.

### 6.9 Manual dogfood
- [x] See "Manual dogfood" block above. Before: 41 stale processes accumulated
  over 4 days. After: count stays at 0 across 5 close/kill iterations.

## Files

- `src/utils/parent-watcher.ts` — new helper (DI'd for testability).
- `tests/utils/parent-watcher.test.ts` — 8 unit tests.
- `src/index.ts` — install + teardown wiring inside the `serve` action.
- `tests/integration/stdio-orphan.test.ts` — end-to-end orphan reproduction.
- `README.md` — env-var table addition.
- `docs/releases/v1.10.2.md` — staging release notes calling out scope and
  no-change paths.

## Out of scope (per issue #644 §4)

- Single-instance lock per port (was P1) — explicitly dropped: would break
  the canonical multi-MCP-client topology where Claude/Codex/IDE plugins
  each spawn their own stdio openchrome against one shared Chrome.
- Idle timeout (P2) and `oc cleanup` CLI (P3) — opt-in / safe defaults
  preferred, deferring to a follow-up PR if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -535,6 +535,8 @@ docker run openchrome
 | `CHROME_USER_DATA_DIR` | Custom profile directory |
 | `CI` | Detected automatically; adds `--no-sandbox` |
 | `DOCKER` | Detected automatically; adds `--no-sandbox` |
+| `OPENCHROME_PPID_WATCH` | Set to `0` to disable the parent-process watcher in stdio mode. Default: enabled. The watcher exits the server when its launching MCP-client parent dies, so abrupt parent termination (force-quit, `kill -9`, tmux teardown) does not orphan the openchrome process. HTTP and `both` transport modes ignore this flag — they remain daemon-capable. |
+| `OPENCHROME_PPID_WATCH_INTERVAL_MS` | Polling interval for the parent watcher in milliseconds. Default: `2000`. Clamped to `[500, 60000]`. |
 
 ### Individual flags
 
--- a/docs/releases/v1.10.2.md
+++ b/docs/releases/v1.10.2.md
@@ -1,0 +1,111 @@
+# OpenChrome v1.10.2 (staging)
+
+Release date: TBD
+
+## Summary
+
+v1.10.2 is a small **patch release** focused on closing the still-open half
+of the v1.10.1 abrupt-parent-death cleanup work.
+
+v1.10.1 hardened the **openchrome → Chrome** direction: when openchrome dies
+abruptly, its launched Chrome tree is now reliably cleaned up via the
+detached `spawnProcessGuardian`. v1.10.2 adds the symmetric
+**MCP-client → openchrome** direction: when the launching MCP-client chain
+(claude / codex / OMC tmux wrapper / IDE host) dies abruptly without
+closing the stdio pipe, the openchrome stdio server now exits on its own
+within a couple of seconds instead of orphaning indefinitely.
+
+This closes the leak documented in
+[issue #644](https://github.com/shaun0927/openchrome/issues/644), where a
+single user environment accumulated 41 live `openchrome-mcp` processes
+across 4 days of MCP session activity.
+
+---
+
+## What changed
+
+### New: parent-process watcher (stdio mode only)
+
+A small polling watcher (`src/utils/parent-watcher.ts`) is installed at the
+end of the `serve` action when the active transport is **stdio**. It records
+`process.ppid` at startup and exits the process via `process.exit(0)` if
+that pid is no longer alive on a subsequent poll.
+
+Operationally:
+
+- enabled by default in stdio mode
+- skipped when the process is already orphaned at startup (`ppid <= 1`)
+- skipped entirely in `--transport http` and `--transport both` (those
+  modes are intentionally daemon-capable and must survive their launching
+  shells)
+- the existing synchronous `process.on('exit')` Chrome cleanup hook still
+  takes Chrome down with it, so the symmetric guardian path is preserved
+
+### Existing v1.10.1 paths are unchanged
+
+This release does **not** modify:
+
+- `spawnProcessGuardian` (still cleans Chrome when openchrome dies)
+- `pid-manager.writePidFile` (still appends; multi-instance per port is
+  intentional)
+- HTTP / `both` transport behavior
+- the cooperative stdin-EOF defense in `src/transports/stdio.ts`
+
+The watcher is purely additive.
+
+---
+
+## Configuration
+
+| Env var | Default | Description |
+|---|---|---|
+| `OPENCHROME_PPID_WATCH` | `1` | Set to `0` to disable the watcher in stdio mode. |
+| `OPENCHROME_PPID_WATCH_INTERVAL_MS` | `2000` | Poll interval in ms. Clamped to `[500, 60000]`. |
+
+No CLI flag changes. No new dependencies.
+
+---
+
+## Operator-impact summary
+
+- Closing a terminal that hosted an MCP client now also terminates its
+  openchrome stdio server within ~2 seconds. Previously the server (and
+  its auto-launched Chrome) would survive indefinitely.
+- HTTP-daemon deployments are unchanged.
+- Existing graceful shutdown paths (SIGTERM/SIGINT) keep working — the
+  watcher is torn down at the very top of `enhancedShutdown` so it
+  cannot fire during the async shutdown chain.
+
+---
+
+## Compatibility notes
+
+v1.10.2 is intended as a backward-compatible patch release.
+
+The only externally observable difference is the new lifecycle behavior
+above. Operators who depend on stdio-launched openchrome surviving its
+parent shell can opt out with `OPENCHROME_PPID_WATCH=0`, but the
+recommended path for that use case is `--transport http` (or
+`--transport both`), where the watcher is never installed.
+
+---
+
+## Verification performed for v1.10.2
+
+- new unit suite: `tests/utils/parent-watcher.test.ts` (8 tests)
+- new integration suite: `tests/integration/stdio-orphan.test.ts`
+  (1 end-to-end test that builds the real stdio server, pins its stdin
+  to a FIFO held open by an out-of-tree keeper, runs it under an
+  intermediate bash, SIGKILLs the bash, and asserts the server exits
+  within 5 seconds — the test fails fast if the new "Parent watcher:
+  enabled" log line never appears, so a startup crash cannot pass as a
+  watcher success)
+- existing test matrix passes unchanged
+
+---
+
+## References
+
+- issue: https://github.com/shaun0927/openchrome/issues/644
+- previous direction: PR #642 (`Kill managed Chrome after abrupt parent death`)
+- previous release: docs/releases/v1.10.1.md
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,6 +14,7 @@ import { createTransport } from './transports/index';
 import { getGlobalConfig, setGlobalConfig } from './config/global';
 import { ToolTier } from './config/tool-tiers';
 import { writePidFile, cleanOrphanedChromeProcesses } from './utils/pid-manager';
+import { installParentWatcher, ParentWatcherHandle } from './utils/parent-watcher';
 import { getVersion } from './version';
 import { ChromeProcessWatchdog } from './chrome/process-watchdog';
 import { TabHealthMonitor } from './cdp/tab-health-monitor';
@@ -325,6 +326,29 @@ program
       console.error('[openchrome] STDIO transport enabled');
     }
 
+    // Parent-process death watcher (issue #644).
+    //
+    // Symmetric to spawnProcessGuardian (which kills Chrome when openchrome
+    // dies). When the launching MCP-client chain (claude/codex/IDE host) is
+    // killed without closing the stdio pipe, stdin EOF never fires and this
+    // server orphans. The PPID watcher polls the parent and exits cleanly
+    // when it disappears, so the existing process.on('exit') hook below can
+    // take Chrome down with it.
+    //
+    // Stdio mode only — HTTP and "both" modes are intentionally daemon-
+    // capable and must survive their launching shells.
+    let parentWatcher: ParentWatcherHandle | null = null;
+    if (transportMode === 'stdio' && process.env.OPENCHROME_PPID_WATCH !== '0') {
+      const parentPid = process.ppid;
+      if (parentPid > 1) {
+        const intervalMs = parseInt(process.env.OPENCHROME_PPID_WATCH_INTERVAL_MS || '', 10) || 2000;
+        parentWatcher = installParentWatcher({ parentPid, intervalMs });
+        console.error(`[openchrome] Parent watcher: enabled (ppid=${parentPid}, interval=${intervalMs}ms)`);
+      } else {
+        console.error('[openchrome] Parent watcher: skipped (already orphaned, ppid<=1)');
+      }
+    }
+
     // ─── Self-Healing Module Wiring (#354) ──────────────────────────────────
 
     const launcher = getChromeLauncher();
@@ -562,6 +586,10 @@ program
     // Update shutdown handler to include self-healing cleanup
     const originalShutdown = shutdown;
     const enhancedShutdown = async (signal: string) => {
+      // Stop the parent watcher first so it cannot fire process.exit during
+      // the async shutdown work below (issue #644).
+      parentWatcher?.stop();
+      parentWatcher = null;
       processWatchdog.stop();
       tabHealthMonitor.stopAll();
       eventLoopMonitor.stop();
--- a/src/index.ts
+++ b/src/index.ts
@@ -341,9 +341,13 @@ program
     if (transportMode === 'stdio' && process.env.OPENCHROME_PPID_WATCH !== '0') {
       const parentPid = process.ppid;
       if (parentPid > 1) {
-        const intervalMs = parseInt(process.env.OPENCHROME_PPID_WATCH_INTERVAL_MS || '', 10) || 2000;
+        // Forward the parsed value when present; otherwise let installParentWatcher
+        // own the default. clampInterval (in parent-watcher.ts) is the single
+        // source of truth for both the default (2000ms) and the [500, 60000] bounds.
+        const rawInterval = parseInt(process.env.OPENCHROME_PPID_WATCH_INTERVAL_MS || '', 10);
+        const intervalMs = Number.isFinite(rawInterval) ? rawInterval : undefined;
         parentWatcher = installParentWatcher({ parentPid, intervalMs });
-        console.error(`[openchrome] Parent watcher: enabled (ppid=${parentPid}, interval=${intervalMs}ms)`);
+        console.error(`[openchrome] Parent watcher: enabled (ppid=${parentPid}, interval=${intervalMs ?? 2000}ms)`);
       } else {
         console.error('[openchrome] Parent watcher: skipped (already orphaned, ppid<=1)');
       }
--- a/src/utils/parent-watcher.ts
+++ b/src/utils/parent-watcher.ts
@@ -60,8 +60,16 @@ export function installParentWatcher(opts: ParentWatcherOptions): ParentWatcherH
   const exit = opts.exitFn ?? ((code: number) => process.exit(code));
   const log = opts.logger ?? ((msg: string) => console.error(msg));
 
+  // `stopped` closes the race where the timer callback is already queued on
+  // the macrotask queue when stop() runs. clearInterval cancels future ticks
+  // but does not dequeue an already-pending callback, so without this flag
+  // the watcher could still call exit(0) mid-shutdown after stop() returned.
+  let stopped = false;
+
   const timer = setInterval(() => {
+    if (stopped) return;
     if (isAlive(opts.parentPid)) return;
+    stopped = true;
     log(`[openchrome] parent pid ${opts.parentPid} is gone, exiting`);
     clearInterval(timer);
     exit(0);
@@ -73,6 +81,9 @@ export function installParentWatcher(opts: ParentWatcherOptions): ParentWatcherH
   }
 
   return {
-    stop: () => clearInterval(timer),
+    stop: () => {
+      stopped = true;
+      clearInterval(timer);
+    },
   };
 }
--- a/src/utils/parent-watcher.ts
+++ b/src/utils/parent-watcher.ts
@@ -1,0 +1,78 @@
+/**
+ * Parent process liveness watcher.
+ *
+ * Symmetric counterpart to spawnProcessGuardian (src/utils/process-guardian.ts):
+ * spawnProcessGuardian kills Chrome when openchrome dies; installParentWatcher
+ * kills openchrome when its launching MCP-client parent dies.
+ *
+ * Wired only in stdio transport mode — HTTP and "both" modes are intentionally
+ * daemon-capable and must survive their launching shells.
+ *
+ * See https://github.com/shaun0927/openchrome/issues/644 for the leak this
+ * complements the cooperative stdin-EOF defense in src/transports/stdio.ts.
+ */
+
+const MIN_INTERVAL_MS = 500;
+const MAX_INTERVAL_MS = 60_000;
+
+export interface ParentWatcherOptions {
+  parentPid: number;
+  intervalMs?: number;
+  logger?: (msg: string) => void;
+  // Injected for unit testing — production callers should not pass these.
+  isAliveFn?: (pid: number) => boolean;
+  exitFn?: (code: number) => void;
+}
+
+export interface ParentWatcherHandle {
+  stop: () => void;
+}
+
+function defaultIsAlive(pid: number): boolean {
+  try {
+    process.kill(pid, 0);
+    return true;
+  } catch (err) {
+    // EPERM = process exists but we cannot signal it (typical on Windows).
+    if ((err as NodeJS.ErrnoException).code === 'EPERM') return true;
+    return false;
+  }
+}
+
+function clampInterval(value: number | undefined): number {
+  const raw = Number.isFinite(value) ? Number(value) : 2_000;
+  if (raw < MIN_INTERVAL_MS) return MIN_INTERVAL_MS;
+  if (raw > MAX_INTERVAL_MS) return MAX_INTERVAL_MS;
+  return raw;
+}
+
+/**
+ * Install a polling watcher that exits this process when `parentPid` is no
+ * longer alive. Returns a handle whose `stop()` cancels the watcher — call
+ * this from the graceful shutdown path so it cannot fire mid-shutdown.
+ *
+ * Caller is responsible for deciding whether to install at all (transport
+ * mode, env var opt-out, ppid > 1 check, etc.).
+ */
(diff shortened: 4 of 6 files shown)


shaun0927:
**CI status note**: this PR's red matrix on the cross-OS jobs is **inherited from `develop`**, not introduced here. The 36 failures match `develop` HEAD (commit 32d8edd) one-for-one — verified by checking out `upstream/develop` clean and running the same `npx jest --config jest.ci.config.js` command, which produces the same set.

The fix is in #646, which cherry-picks two missing commits already on `main` (#643 hotfix + the test-hardening commit `59be0ad`). Once #646 lands on `develop`, please rebase or merge `develop` into this branch and re-trigger CI — local verification with both branches stacked shows 1 failed (a known timing flake on `event-loop-monitor`) and 3206 passed, with 9/9 green when the flake is rerun in isolation.

Order:
1. Merge #646 into `develop` (no functional changes, just CI plumbing).
2. Rebase / merge `develop` into this branch.
3. Re-run CI. Expect 9/9 green.

The actual P0 work in this PR — `tests/utils/parent-watcher.test.ts` (8 tests) and `tests/integration/stdio-orphan.test.ts` (1 test) — passes locally on this exact tree.

shaun0927:
**Review feedback addressed in commit 85274c8.**

Phase-4 multi-perspective review (architect, security, code-reviewer) returned APPROVE × 3 with two MEDIUM correctness items from the code-reviewer:

1. **Race window between queued timer callback and `stop()`** — `clearInterval` cancels future ticks but does not dequeue an already-pending macrotask. Without an internal `stopped` flag, a callback already on the queue when `enhancedShutdown` calls `parentWatcher.stop()` could still fire `process.exit(0)` mid-shutdown. Added the flag in `src/utils/parent-watcher.ts` and short-circuit at the top of the callback. Also flip the flag *inside* the dead-parent branch so the exact-once guarantee holds even without `clearInterval`.

2. **Double defaulting of `intervalMs`** — both the `serve` action and `clampInterval` provided a 2000ms default, creating two sources of truth and an off-by-fence behavior on negative env values. Removed the fallback in `src/index.ts` and let `clampInterval` own both the default and the `[500, 60000]` bounds. Forward `undefined` when the env var is unset/non-numeric.

Plus a new unit test for case (1) covering 'stop() prevents exit even if a queued callback observes a dead parent'. Suite is now 9/9 green.

`npm run lint` 0 errors. `npm run build` clean. `npx jest tests/utils/parent-watcher.test.ts tests/integration/stdio-orphan.test.ts` is 10/10 green on the rebuilt dist.

Architect verdict: APPROVE (zero blocking, one P2 about `process.exit` skipping async cleanup — accepted as deliberate trade-off for orphan scenario, documented in v1.10.2 release notes).
Security verdict: APPROVE (no new attack surface, env-var clamping correct, integration test paths derived from `os.tmpdir()` + `mkdtempSync`, `process.kill(pid, 0)` is non-injectable).
Code-reviewer verdict: COMMENT → now APPROVE-ready after the two MEDIUM fixes.

shaun0927:
**🟢 9/9 CI matrix green** after rebase onto develop (post-#646 merge).

| OS | Node 18 | Node 20 | Node 22 |
|---|---|---|---|
| ubuntu-latest | ✅ | ✅ | ✅ |
| macos-latest | ✅ | ✅ | ✅ |
| windows-latest | ✅ | ✅ | ✅ |

Same 9-job matrix that v1.10.1 documented as the green baseline. Ready for review/merge.

[Run 24814867168](https://github.com/shaun0927/openchrome/actions/runs/24814867168)

PR #653, opened 2026-04-27
perf: token-efficiency pass — strip decorators, add validate_page macro

marcosremar:
## Summary

Five focused changes that cut token usage in a typical "validate a frontend page" workflow from ~4000 tokens (8-12 round-trips) to ~550 tokens (1 round-trip), and capture load-time console errors that the multi-call pattern silently dropped.

### Changes

1. **Promote \`console_capture\` to Tier 1.** Was Tier 2 — agents fell back to \`javascript_tool\` polling \`window.__check()\` because they couldn't see the proper tool. (\`src/config/tool-tiers.ts\`)

2. **Skip \`_sessionContext\` and \`_profile\` decorators on read-only tools** (\`read_page\`, \`query_dom\`, \`find\`, \`inspect\`, \`javascript_tool\`, \`wait_for\`, \`page_content\`, \`tabs_context\`). These tools fire 50× per session and don't change session state — resumability metadata isn't load-bearing here. (\`src/mcp-server.ts\`)

3. **Drop the duplicate hint from \`content[]\`.** The hint was already delivered via \`_hint\` and \`_hintMeta\`; the \`content.push({...})\` block was shipping it a second time. (\`src/mcp-server.ts\`)

4. **\`wait_for\` description discourages \`type=timeout\` polling.** Strongly recommend \`function\`/\`selector\`/\`url_match\` which return as soon as the condition is true (1 round-trip) instead of sleeping for a fixed duration and forcing a follow-up read. (\`src/tools/wait-for.ts\`)

5. **New \`validate_page\` macro tool.** One call: navigate → wait for ready → capture console errors via CDP \`Runtime.consoleAPICalled\` (so load-time errors land — not just post-load) → return \`{tabId, url, title, status, console:{errors,warnings,totals}, summary:{interactiveCount, formCount, hasCanvas, hasIframe, bodyTextSample}}\`. Designed for "is this page healthy?" workflows that previously required \`navigate\` + \`wait_for\` + \`console_capture start\`/\`get\` + \`read_page\` + \`evaluate\`. (\`src/tools/validate-page.ts\`)

### Real-world impact

Tested locally against \`http://localhost:5183/admin/scenery\` (a Vite + React + Babylon.js admin):

| | Before | After |
|---|---|---|
| Tokens per page validation | ~4000 | **~550** |
| Round-trips | 8-12 | **1** |
| Wall time | ~30s | **~10s** |
| Load-time error capture | dropped | **captured** |

The headline catch in the local run: a \`WebGL not supported\` thrown during Babylon.js \`_ThinEngine\` init — emitted before any user-installed \`window.__cmConsoleErrors\` collector could be evaluated. The macro's CDP listener is attached BEFORE navigation, so it lands.

## Test plan

- [x] \`npm install && npm run build\` clean
- [x] \`npm test\` — 3664 pass, 0 fail (one pre-existing test asserting Tier 1 count of 31 was updated to 33 to reflect \`console_capture\` and \`validate_page\` being promoted)
- [x] Local stdio JSON-RPC smoke test calls \`validate_page\` against a real Vite dev server, confirms 65 tools registered (was 63), and the response matches the documented shape
- [ ] Reviewer eyeballs the new tool description / schema

### DIFF
--- a/src/config/tool-tiers.ts
+++ b/src/config/tool-tiers.ts
@@ -41,6 +41,7 @@ export const TOOL_TIERS: Record<string, ToolTier> = {
   oc_copy_to_clipboard: 1,
   oc_open_host_settings: 1,
   act: 1,                     // src/tools/act.ts — composite NL action API (#578)
+  validate_page: 1,           // src/tools/validate-page.ts — composite page-health check (#token-efficiency)
 
   // Tier 2: Specialist (on demand)
   extract_data: 2,              // src/tools/extract-data.ts — structured extraction (#571)
--- a/src/config/tool-tiers.ts
+++ b/src/config/tool-tiers.ts
@@ -31,6 +31,7 @@ export const TOOL_TIERS: Record<string, ToolTier> = {
   wait_for: 1,
   memory: 1,
   lightweight_scroll: 1,
+  console_capture: 1,         // src/tools/console-capture.ts — surfaces page errors during validation (#token-efficiency)
   oc_stop: 1,
   oc_profile_status: 1,
   oc_session_snapshot: 1,
@@ -54,7 +55,6 @@ export const TOOL_TIERS: Record<string, ToolTier> = {
   page_pdf: 2,
   page_screenshot: 2,
   page_content: 2,
-  console_capture: 2,
   performance_metrics: 2,
   file_upload: 2,
   batch_execute: 2,
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -1266,7 +1266,7 @@ export class MCPServer {
         }
       }
 
-      // Inject proactive hint into both _hint (backward compat) and content[] (guaranteed MCP delivery)
+      // Inject proactive hint into _hint and _hintMeta (content[] duplicate removed for token efficiency #token-efficiency)
       if (this.hintEngine) {
         const hintResult = this.hintEngine.getHint(toolName, result as Record<string, unknown>, false, sessionId);
         if (hintResult) {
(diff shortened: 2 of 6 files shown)


shaun0927 (commented):
Thanks @marcosremar — really nice work. The `validate_page` macro and the `wait_for` guidance are clear wins, and promoting `console_capture` to Tier 1 is the right call.

Plan from the maintainer side: I'd like to land this on `develop` with two small follow-up patches on top, rather than asking you to round-trip:

1. **Tighten `READONLY_HIGH_FREQ_TOOLS` in `src/mcp-server.ts`.**
   `javascript_tool` can mutate page/session state (it executes arbitrary JS), so it shouldn't share the same decorator-skip path as truly read-only tools like `read_page` / `query_dom` / `inspect`. I'll drop it from the set and rename the constant to `STATE_STABLE_HIGH_FREQ_TOOLS` for accuracy.

2. **Restore symmetry between the success and error hint paths.**
   The original `content.push({type:'text', text: hint})` was intentional — the inline comment called it _"guaranteed MCP delivery"_ because `_hint` / `_hintMeta` are non-standard fields some clients won't surface. The error path still pushes to `content[]`; we should either keep both or remove both. I'll keep the `content[]` push for parity, since the per-call token cost is small and the cross-client visibility guarantee matters more.

Everything else (`validate_page`, `wait_for` description, `console_capture` tier promotion, test count update) lands as-is. Authorship on your commits will be preserved via the merge.

Also: CI is currently in `action_required` state because this is your first contribution — I'll approve the workflow run before merging so we get a green build first.

PR #654, opened 2026-04-27
feat(computer): add screenshotFormat option to override webp default

tommymarshall:
## Summary

The `computer` tool's `screenshot` action is hardcoded to webp, which is great for token-budget-conscious agents but unusable in MCP clients that cannot decode `image/webp` inline. Cursor's `CallMcpTool`, for example, surfaces every webp screenshot as:

```
Mime type image/webp does not support decoding
```

regardless of `screenshotQuality` (`high`, `normal`, `low`) or `force: true`. Today the only escape hatches are:

- `expand_tools(tier="2")` then `page_screenshot` with `format: "png"` (works, but requires re-entering Tier 2 every session and paying a per-tab tool-list cost), or
- attaching a CDP client to `127.0.0.1:9222` and calling `Page.captureScreenshot` with `format: "png"` directly (works, but requires the agent to roll its own WebSocket handshake — and most off-the-shelf clients trip the `--remote-allow-origins` check on their first try).

This PR adds a `screenshotFormat` argument to the existing `computer` tool so agents can ask for PNG (or JPEG) via the same Tier 1 entry point they already use.

## Behavior

- New optional `screenshotFormat` arg on `computer`: `"webp" | "png" | "jpeg"`.
- Default stays `"webp"` — no change for existing callers, no payload increase by surprise.
- `"png"` is lossless: the CDP call omits `quality` and `optimizeForSpeed` (both only apply to lossy formats and CDP rejects `quality` for PNG).
- `"jpeg"` reuses the existing `screenshotQuality` preset map and `optimizeForSpeed: true`, parallel to webp.
- Unrecognized values fall back to `"webp"` rather than erroring; the schema enum keeps spec-compliant clients honest.
- Adaptive screenshot, retry-once, DOM fallback, force/quality, and text-only modes are all untouched.

## Why not just promote `page_screenshot` to Tier 1?

`page_screenshot` is broader (file output, full-page, clip regions, omitBackground) and intentionally Tier 2. Adding a single `screenshotFormat` arg to `computer` is the minimal change that fixes the inline-render mime-type problem without reshuffling tiers or duplicating capture logic.

## Test plan

- [x] `npx jest tests/tools/computer.test.ts` — all 109 tests pass, including 3 new ones:
  - `'png' format returns image/png and omits quality / optimizeForSpeed`
  - `'jpeg' format returns image/jpeg with quality set`
  - `unrecognized format falls back to webp`
- [x] Existing webp screenshot tests unchanged and still pass (default behavior preserved).
- [x] `npm run build` clean.
- [x] `npm run lint` — 0 errors (44 pre-existing warnings, none in `computer.ts`).
- [x] `npm test` — 3582 passed, 86 skipped (pre-existing skips).

## Real-world repro

Used this build to seed an Auth0-protected SPA in Cursor's MCP browser via openchrome (login + localStorage seed + reload + verify). Before this change, `computer.screenshot` returned `Mime type image/webp does not support decoding` and we had to capture PNG via a raw CDP WebSocket. With `screenshotFormat: "png"`, the same call renders inline.

## Notes for reviewer

- Schema description for `screenshotFormat` calls out the use case explicitly so agents pick PNG when needed and don't pay the bigger payload by default.
- I left the variable named `screenshotFormat` (parallel to `screenshotQuality`) rather than just `format` to avoid colliding with future per-action format arguments on other actions in the same tool.
- Happy to reorder to `format` if you'd prefer; just let me know.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,6 +409,48 @@ Tier 3 launches a real headed Chrome window with a genuine user-agent (`Chrome/.
 
 ---
 
+## FAQ: Comparison with Other Browser MCPs
+
+Common questions from users evaluating OpenChrome against Chrome DevTools MCP, Firefox DevTools MCP, and similar tools (see [#612](https://github.com/shaun0927/openchrome/issues/612)).
+
+### Can multiple MCP clients share tabs safely?
+
+**Yes — tabs cannot clobber each other across clients.**
+
+OpenChrome identifies every tab by its CDP `targetId` — a stable, browser-assigned string — not by a visible 1/2/3 index. On top of stable IDs, two layers of isolation are specifically designed for multi-client scenarios:
+
+- **`workerId`** — logical tab groups per client or parallel lane. A new tab under one `workerId` is invisible to another and never replaces one.
+- **`profileDirectory`** — launches a fully separate Chrome instance bound to that profile, giving OS-level cookie / storage / extension isolation.
+
+If client A opens five tabs and client B opens five tabs, all ten `tabId`s are distinct and stable; a new tab from A can never displace B's tab #3.
+
+### How do I handle sites that require interactive login (password, 2FA, CAPTCHA)?
+
+Two mechanisms, used together in practice:
+
+**1. Persistent-profile headless — log in once, automate forever.**
+Point OpenChrome at a persistent `userDataDir` (+ optional `profileDirectory`) and cookies / `localStorage` / IndexedDB survive across runs. After the first login, subsequent **headless** runs stay logged in until the site invalidates the session.
+
+**2. Headed fallback for the initial login or WAF-blocked sites.**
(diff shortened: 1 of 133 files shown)


tommymarshall:
@shaun0927 let me know if you have any questions :+1:

shaun0927:
Thanks for the well-scoped PR @tommymarshall — the problem statement, the CDP-correct PNG handling (omitting `quality`/`optimizeForSpeed` for lossless), the webp-default backward-compat, and the test coverage are all spot-on.

One housekeeping item: this repo merges contributions into `develop` first and then promotes `develop` → `main` on release. I've retargeted this PR to `develop` for you (no diff changes — `src/tools/computer.ts` and `tests/tools/computer.test.ts` are identical between the two branches right now).

Re: the `screenshotFormat` vs `format` naming question in your PR description — let's keep `screenshotFormat`. Your reasoning (avoiding collision with future per-action format args on other actions in the same tool) is correct.

I'll let CI run on the new base and merge once it's green.

Issue #660, opened 2026-05-01
Close button on launched Chrome window is ignored — process auto-relaunches


shaun0927:
## Summary

When openchrome launches Chrome and the user clicks the OS close button (macOS red dot / Windows X / Linux WM close), the window closes — but the MCP server **silently re-launches Chrome within ~10 seconds** via the watchdog. Users cannot stop the spawned browser without calling `oc_stop` (which most users don't know about) or killing the parent agent session.

## Symptoms

1. Run `node dist/index.js --auto-launch`.
2. After Chrome appears, click the close button.
3. Wait 10–15 seconds.
4. Chrome reappears with a blank tab.
5. Repeat: same thing happens.

### 1. Detached spawn, no signal back-channel
`src/chrome/launcher.ts:605-609`:

```ts
const chromeProcess = spawn(chromePath, args, {
  detached: true,
  stdio: ['ignore', 'ignore', 'pipe'],
});
chromeProcess.unref();
```

`detached: true` puts Chrome in its own process group; the OS terminates Chrome cleanly when the user clicks close, but no IPC channel back to the MCP server signals "user intent". The exit handler at `:634-642` only logs.

### 2. Watchdog blindly re-launches
`src/chrome/process-watchdog.ts:130-149` polls every 10s (60s when idle) and calls `ensureChrome({ autoLaunch: true })` whenever it sees a dead Chrome PID, with no concept of "user closed it intentionally vs. crashed".

### 3. CDP client reconnect loop
`src/cdp/client.ts:560-628` has an unbounded reconnect loop; while it uses `autoLaunch: false`, the watchdog (separate timer) launches Chrome anyway.

### 4. `_intentionalStop` flag is too narrow
`src/chrome/launcher.ts:316-318, 742` — `_intentionalStop` is set only when `oc_stop` (or MCP shutdown) explicitly calls `launcher.close()`. User-driven OS close never sets it.

## Design Constraints

1. **Don't break legitimate auto-recovery**: Chrome legitimately crashes (renderer OOM, segfault, OOM-killer). Auto-recovery is valuable for long-running scrape jobs.
2. **No site-specific logic**.
3. **Default behavior must match user intuition**: closing a window means "stop, please". Power users who actually want auto-recover for crash-resilience should opt in.
4. **Cross-platform**: macOS (close → SIGTERM, exit code 0), Windows (WM_CLOSE → exit code 0), Linux (WM close → SIGTERM exit code 0). All three look the same to Node's `exit` event.

[...]

Issue #659, opened 2026-05-01
Profile selection opens new isolated window instead of attaching to user's running Chrome


shaun0927:
## Summary

When the user specifies a Chrome profile, openchrome spawns a **new Chrome process** with an **isolated `--user-data-dir`** instead of attaching to the user's already-running Chrome. The spawned window has no logins, no cookies, no extensions — even though the user is fully signed in to the same profile in their daily Chrome on the same machine.

Desired behavior: open a new tab in the user's existing Chrome with all logins intact.

## Symptoms

1. User runs Chrome day-to-day, signed into Google/GitHub/etc. on `Profile 1`.
2. Agent runs an openchrome tool with `profileDirectory: "Profile 1"`.
3. openchrome opens a fresh Chrome window in `~/.openchrome/profiles/Profile_1/` — empty cookies, no extensions.
4. User has to log in again inside the spawned window.

### 1. Chrome `SingletonLock` is per-`--user-data-dir`
Two Chrome processes cannot share the same `--user-data-dir`. To avoid that conflict, `src/chrome/pool.ts:355` creates a **new isolated** dir per profile:

```ts
const profileUserDataDir = profileDirectory
  ? path.join(os.homedir(), '.openchrome', 'profiles', profileDirectory.replace(...))
  : undefined;
```

Even though `--profile-directory` matches the user's profile name, the `--user-data-dir` is different — so Chrome treats it as a brand-new clone with empty state.

### 2. Always-spawn model
`src/chrome/launcher.ts:605` always calls `spawn()`. There IS attach logic at `src/chrome/launcher.ts:366-378` (HTTP probe to `/json/version` on `--remote-debugging-port`), but it only succeeds if the user's Chrome was already started with `--remote-debugging-port=NNNN`, which normal users never do.

### 3. Chrome 136+ refuses debug-port on real profile
`src/chrome/launcher.ts:513` notes that Chrome 136+ refuses `--remote-debugging-port` against the real Default profile, forcing the copy-fallback in 99% of cases.

### 4. Pool isolation policy
`src/chrome/pool.ts:140-200` always launches a new instance per profile; no "adopt the running Chrome" path.

## Design Constraints

1. **Do not modify the user's Chrome installation or daily-driver state without explicit consent**. Auto-restarting a Chrome with 100 open tabs is unacceptable without a clear opt-in.
2. **Cross-platform**: macOS / Windows / Linux must all work, with platform-specific Chrome paths and process inspection commands.
[...]

Issue #524, opened 2026-03-31
Desktop App: Error handling + local fallback + CLI coexistence


shaun0927:
Parent: #518 | Depends on: #519, #520, #521
Seed: `seed_d1dff08c8404`

## Overview

Implement the error handling philosophy (auto-heal → friendly UI guide → local fallback) and CLI coexistence logic (detect existing server, connect instead of conflict).

## Error Philosophy

```
Error occurs
    │
    ▼
Auto-heal silently (retry, reconnect, restart)
    │
    ├─ Success → user never sees the error
    │
    ▼ Failed
Show friendly UI guidance (no jargon, actionable button)
    │
    ├─ User action resolves → back to normal
    │
    ▼ Unrecoverable
Graceful degradation (local-only mode, limited functionality)
```

### Chrome Not Found
- [x] Detect Chrome on app startup (check standard install paths per platform)
- [x] macOS: `/Applications/Google Chrome.app/Contents/MacOS/Google Chrome`
- [x] Windows: `%PROGRAMFILES%\Google\Chrome\Application\chrome.exe` + `%PROGRAMFILES(X86)%\...`
- [x] Linux: `google-chrome`, `google-chrome-stable`, `chromium-browser` (PATH search)
- [x] If not found: show friendly message in main window with "Download Chrome" button
- [x] "Download Chrome" → opens `https://www.google.com/chrome/` in default browser
- [x] App remains open (don't exit) — user can install Chrome and click "Retry"
- [x] If Chrome found later (retry): auto-detect and enable Start button

### Tunnel Failure Handling
- [x] Tunnel fails to establish (firewall/no internet/antivirus):
  - Show "Local Mode" badge in status bar
  - Disable web host connect buttons (grey out with tooltip "No internet connection")
  - Enable "Connect to Cursor" and "Connect to Claude Desktop" (local stdio still works)
  - Show retry button for tunnel
  - Log reason to app log (not shown to user unless they expand "Details")
- [x] Tunnel drops mid-session:
  - Auto-reconnect immediately (silent, up to 3 attempts over 30 seconds)
  - If network blip < 10 seconds → fully transparent, user sees nothing
  - If extended outage → show "Reconnecting..." indicator with spinner
  - After 3 failed reconnects → fall back to local-only mode
  - When network returns → auto-reestablish tunnel, remove "Local Mode" badge
- [x] cloudflared binary blocked by antivirus:
[...]

PR #667, opened 2026-05-01
feat(#661): synchronous Chrome shutdown + ownership-marker reaper

### LINKED ISSUE #661: Zombie Chrome processes leak across sessions, eating RAM
## Summary

Each Claude Code or Codex session leaves orphaned Chrome processes (and macOS Helper / Renderer / GPU subprocesses) running after the parent agent exits. RAM compounds across sessions until the user manually kills them or reboots — typically several gigabytes after a working day.

This is the highest-impact issue of the five reported because it actively degrades the user's machine over time.

## Symptoms

```bash
# Open Claude Code, do some openchrome work, exit
# Then:
ps aux | grep -i chrome | grep -v grep | wc -l
# Multiple Chrome processes still alive.
# Repeat sessions accumulate.
```

## Root Cause

A lifecycle gap between **stdin EOF** (parent exits) and **Chrome termination**. The MCP server takes a synchronous `process.exit(0)` path that bypasses the async cleanup that would actually kill Chrome.

### 1. Synchronous exit on stdin EOF
`src/transports/stdio.ts:80-91`:

```ts
this.rl.on('close', () => {
  console.error('[StdioTransport] stdin closed (readline), shutting down...');
  process.exit(0);          // async cleanup never awaited
});

process.stdin.on('end', () => {
  console.error('[StdioTransport] stdin ended, shutting down...');
  process.exit(0);
});
```

### 2. Detached, unref'd Chrome
`src/chrome/launcher.ts:605, 622`:

```ts
const chromeProcess = spawn(chromePath, args, { detached: true, stdio: [...] });
chromeProcess.unref();
```

`detached: true` puts Chrome in a fresh process group; `unref()` removes it from Node's reference count. When MCP dies abruptly, OS does not propagate signals to Chrome's group.

### 3. Guardian is also detached
`src/utils/process-guardian.ts:76-77` spawns a polling watchdog that's itself detached. If both parent and guardian die in the same instant (terminal close, sudden reboot), Chrome survives.

### 4. Orphan reaper only runs at next startup
`src/utils/pid-manager.ts:173-204` `cleanOrphanedChromeProcesses()` runs only on clean startup. It matches by PID file (`/tmp/openchrome-chrome-{port}.pid`); next startup with a different port misses prior orphans.

### 5. macOS Helper subprocesses
`Google Chrome Helper (Renderer/GPU/Plugin)` spawn per tab. There is no explicit cascade kill.

## Design Constraints

1. **Never kill the user's daily-driver Chrome by mistake**. The user may have their own Chrome running outside openchrome; orphan-reaping must be precisely scoped to processes openchrome itself launched.
2. **No site / no per-tool logic**: the fix is in lifecycle plumbing.
3. **Cross-platform**: macOS, Linux, Windows.
4. **Coexist with Issue #659 attach mode**: when openchrome attaches to a user's running Chrome, that process must NEVER be killed on exit. Attach mode requires a different cleanup discipline (close attached connections only, leave Chrome alive).
5. **Survive ungraceful parent death**: SIGKILL of Claude/Codex, OS reboot, terminal-close-without-exit-handler.
6. **Preserve `oc_session_resume`**: long-running session resume legitimately needs Chrome to outlive a brief MCP restart. Provide an opt-out via session-resume tokens (defined below).
7. **Coordinate with Issue #660 quiesce**: a Chrome that the user closed (clean exit, watchdog quiescing) is still owned by an alive MCP parent; the reaper must NOT touch it just because Chrome is currently absent.

## Proposed Solution

A defense-in-depth approach. No single change suffices.

### Phase 1 — Ownership marker file (foundational)

When openchrome launches a Chrome process *that it owns*, write a marker file inside that Chrome's `--user-data-dir`:

```
<user-data-dir>/.openchrome-managed
```

Content (JSON):
```json
{
  "pid": 12345,
  "ppid": 67890,
  "ppidCommand": "/usr/local/bin/node /path/to/openchrome/dist/index.js --auto-launch",
  "userDataDir": "/Users/me/.openchrome/profiles/Profile_1",
  "startedAt": "2026-05-01T03:00:00.000Z",
  "marker": "<random-uuid>",
  "launchMode": "isolated"
}
```

This marker:
- Distinguishes "Chrome we launched" from "user's Chrome we attached to" (issue #659 attach mode never writes a marker).
- Is the source of truth for orphan-reaping (only kill processes whose user-data-dir contains a marker AND whose marker UUID matches).
- Stores `userDataDir` explicitly so the fallback marker location (below) can still be cross-checked.
- Stores `ppidCommand` so the reaper can verify the parent PID still belongs to openchrome (PID-reuse defense).

**Fallback marker location**: if the profile dir is unwritable, write to `~/.openchrome/state/markers/<pid>.json` instead. The reaper scans both locations.

### Phase 2 — Synchronous best-effort kill on exit

In `src/transports/stdio.ts`, replace each `process.exit(0)` with:

```ts
this.rl.on('close', () => {
  shutdownSyncBestEffort();
  process.exit(0);
});
```

`shutdownSyncBestEffort()` (new, in `src/transports/stdio.ts` or a sibling):
1. Skip entirely if `OPENCHROME_KILL_ON_EXIT === 'never'` OR if an active session-resume token exists (see Phase 7).
2. Read the launcher's tracked Chrome PID(s) AND launchMode from a synchronously-readable module-level variable. **Skip any instance with `launchMode === 'attach'`** (Issue #659 coordination).
3. For each managed PID, call `killProcessTree(pid, 'SIGTERM')` (Phase 4).
4. After a 200ms grace, call `killProcessTree(pid, 'SIGKILL')` for any survivors.
5. Remove marker files.
6. `process.exit(0)`.

This costs ~200ms but reliably terminates Chrome on graceful parent exit.

### Phase 3 — `process.on('exit')` synchronous handler

Even though `process.exit()` skips async handlers, **synchronous** `'exit'` listeners do run. Register one that calls `shutdownSyncBestEffort()` as a belt-and-suspenders second line of defense (handles cases like uncaught exceptions that bypass stdio transport handlers).

### Phase 4 — Cross-platform process-tree kill

`killProcessTree(pid, signal)` in `src/utils/process-tree.ts` (new). Implementation order of preference:

1. **Preferred (POSIX)**: `process.kill(-pid, signal)` (note the leading minus → signals the entire process group). This works because the launcher spawned Chrome with `detached: true`, which makes `pid` the PGID.
2. **Fallback (POSIX)**: if step 1 fails (`ESRCH` / `EPERM`), recursively walk children via `pgrep -P <pid>` and signal each leaf-up.
3. **Windows**: `taskkill /PID <pid> /T /F` — `/T` cascades to descendants, `/F` forces. One sync call.

All shell-outs use `child_process.spawnSync` with explicit timeouts (1s) so the synchronous shutdown path stays bounded.

### Phase 5 — Aggressive orphan reaper at startup

Replace `cleanOrphanedChromeProcesses()`:
1. Scan `~/.openchrome/profiles/*/.openchrome-managed` and `~/.openchrome/state/markers/*.json` for orphans.
2. For each marker, run a four-way check before killing:
   - **Alive check**: `process.kill(pid, 0)` — process exists?
   - **Identity check**: read the running process's command line and confirm `--user-data-dir` matches the marker's `userDataDir`. Platform-specific:
     - Linux: `/proc/<pid>/cmdline` (NUL-separated, never truncated).
     - macOS: `ps -ww -o command= -p <pid>` (`ww` flag avoids the default ~1024-char truncation; without it, long Chrome command lines lose the end). If `--user-data-dir` is not in the output, treat as ambiguous and DO NOT kill.
     - Windows: `wmic process where ProcessId=<pid> get CommandLine /format:list` or `Get-CimInstance Win32_Process`.
   - **Parent-alive check**: read marker's `ppid`. If alive AND its command line matches `ppidCommand` (still openchrome), DO NOT reap — even if Chrome appears idle (this is Issue #660's quiesced-but-managed case). Coordinates with Issue #660.
   - **Marker-mismatch defense**: if any of the above fail to confirm ownership (e.g., PID reuse, truncated cmdline on macOS without `ww`), DO NOT kill; remove only the stale marker file.
3. If all four checks pass and the parent is dead/reused, `killProcessTree(pid, 'SIGTERM')` (5s grace) → `'SIGKILL'`, then remove marker.
4. Run on **every** startup, not just when a PID file exists.
5. Expose as `oc_reap_orphans` tool for manual sweep.

### Phase 6 — Attach-mode safety (Issue #659 coordination)

In Issue #659's attach mode, do **not** write a marker file. The cleanup paths (Phases 2, 3, 5) only operate on processes with markers. Attached Chrome is never killed.

### Phase 7 — Session-resume token

Define the artifact concretely:
- **Format**: file at `~/.openchrome/state/session-resume-<mcpPid>.json`.
- **Created by**: `oc_session_snapshot` tool (or `oc_checkpoint`) — writes `{ chromePid, port, profileDir, ttlEpochMs }`.
- **Consumed by**: `oc_session_resume` tool — reads the file, verifies TTL, attaches to the alive Chrome by port.
- **TTL**: default 30 minutes; configurable via `OPENCHROME_SESSION_RESUME_TTL_MIN`.
- **Cleanup**: token file is deleted by `oc_session_resume` on success, or by the reaper if `Date.now() > ttlEpochMs` (token-expired path).

`OPENCHROME_KILL_ON_EXIT` mode semantics:
- `auto` (default): Phase 2 sync-kill skips iff a non-expired token exists for this MCP PID.
- `always`: ignore tokens, always sync-kill.
- `never`: never sync-kill. Reaper still runs at next startup.

### Phase 8 — Reconcile with `detached: true`

Keep `detached: true`. The detachment is no longer a problem because:
- Phase 2/3 sync-kill on graceful exit (using PGID via `process.kill(-pid, …)`).
- Phase 5 reaper handles ungraceful exit on next startup.
- Phase 1 marker scopes everything to processes we own.

## Backwards Compatibility

- Tools, APIs, and tool inputs unchanged.
- Behavior change is invisible during normal use; only "I left a session running and exited cleanly" goes from "Chrome leaks" to "Chrome cleans up". This is the desired fix.
- `~/.openchrome/profiles/*` directories now contain a hidden `.openchrome-managed` file. No conflict with existing files.
- Users relying on Chrome surviving MCP exit set `OPENCHROME_KILL_ON_EXIT=never` OR use the session-resume token workflow.
- Quiesced Chrome (Issue #660) — still managed by an alive MCP — is preserved by the parent-alive check.

## Edge Cases & Risks

1. **User has Chrome running with our same `--user-data-dir`** (theoretically impossible due to SingletonLock, but defensive): marker-based scoping means we only kill Chromes we launched.
2. **Reaper runs while a sibling MCP is starting up**: the four-way check (alive + identity + parent-alive + marker-match) prevents misfire.
3. **PID reuse**: between orphan death and next startup, the PID could be reused by an unrelated process. Mitigation: identity check (`--user-data-dir` cross-match) plus marker UUID stored both in the file and embedded in a sentinel cookie (`<user-data-dir>/.openchrome-managed-cookie`) so even if the unrelated process is also Chrome, its cookie won't match.
4. **Headless Chrome leaves no GUI but same process tree**: same fix applies.
5. **Long-running session (`oc_session_resume`)**: handled by Phase 7 token + `OPENCHROME_KILL_ON_EXIT=auto`.
6. **Windows Job Objects**: an alternative to PID-tracking on Windows is creating a Job Object with kill-on-jobclose. Out of scope for this issue but documented as future work.
7. **Filesystem permissions**: marker file writes must not fail loudly; on read-only profile dirs, fall back to `~/.openchrome/state/markers/<pid>.json` (less reliable but better than nothing) — the marker JSON contains `userDataDir` so the reaper can still verify identity.
8. **OOM-killer takes the parent**: kernel sends SIGKILL — no userland chance to clean up. Phase 5 reaper handles on next startup.
9. **Antivirus / sandbox software** kills our marker writes: detect the failure, log, fall back to state-dir markers.
10. **Profile reuse across runs**: same `--user-data-dir` path used in run N and run N+1. Markers are atomic-replaced on launch.
11. **Quiesced Chrome (Issue #660)**: alive MCP, dead Chrome (user closed it). Reaper finds no alive PID → no reap action needed (that's a launcher concern, not the reaper's). If a stale marker remains, parent-alive check spares it from any cleanup.
12. **Attach mode (Issue #659)**: no marker → never reaped. Phase 2 sync-kill explicitly skips `launchMode === 'attach'`.
13. **macOS truncated cmdline**: explicit `ps -ww` flag in Phase 5 identity check. Without `ww`, do NOT kill (safer to leak than to misidentify).

## Verification Procedure

```bash
npm run build

# 1. Graceful exit
node dist/index.js --auto-launch &
PID=$!
sleep 3
# (issue some tool calls)
kill -TERM $PID    # graceful shutdown
sleep 2
ps aux | grep -- '--user-data-dir.*openchrome' | grep -v grep | wc -l
# expect: 0

# 2. Ungraceful exit
node dist/index.js --auto-launch &
PID=$!
sleep 3
kill -KILL $PID    # SIGKILL — bypass shutdown handlers
sleep 2
ps aux | grep -- '--user-data-dir.*openchrome' | grep -v grep | wc -l
# expect: 1+ (orphans expected here, will be reaped on next startup)

# 3. Reaper next startup
node dist/index.js --auto-launch &
PID=$!
sleep 5
ps aux | grep -- '--user-data-dir.*openchrome' | grep -v grep | wc -l
kill -TERM $PID
# expect: only the new Chrome alive, prior orphan killed by reaper

# 4. Cumulative test
for i in {1..10}; do
  node dist/index.js --auto-launch &
  P=$!; sleep 3; kill -TERM $P; sleep 2
done
ps aux | grep -i chrome | grep -v grep | wc -l
# expect: 0 (or just the user's daily-driver Chrome, untouched)

# 5. Daily-driver safety
google-chrome --user-data-dir=/tmp/my-real-chrome &
node dist/index.js --auto-launch &
P=$!; sleep 3; kill -TERM $P; sleep 2
# Verify the user's chrome is still running.
# expect: still alive

# 6. Session-resume preserves Chrome
node dist/index.js --auto-launch &
P=$!; sleep 3
# (in agent) call oc_session_snapshot
kill -TERM $P; sleep 2
# Chrome should still be alive (token preserves it)
ps aux | grep -- '--user-data-dir.*openchrome' | grep -v grep | wc -l
# expect: 1 (the snapshotted Chrome)
# Then start a new MCP, call oc_session_resume — verify reattach succeeds.

# 7. Token-expired cleanup
# Create a session-resume token with ttlEpochMs in the past.
# Start MCP. Reaper runs.
# expect: token file removed, orphan Chrome killed.

# 8. Attach mode (Issue #659)
# Set chromeLaunchMode=attach. Run a tool. Exit MCP.
# expect: user's Chrome still alive; no marker written.

# 9. macOS cmdline truncation regression
# On macOS, force a Chrome with very long --user-data-dir path (~2KB).
# Verify reaper uses `ps -ww` and correctly identifies (or safely refuses to kill).
```

Automated:
- Unit tests for `killProcessTree` per platform (mocked spawnSync). Specifically: PGID path on POSIX, `taskkill /T /F` on Windows.
- Unit tests for marker write/read/verify with fake filesystem (both primary and fallback paths).
- Unit tests for the four-way reaper check, including PID-reuse simulation.
- Integration test that simulates orphan reaping with synthetic markers (different ppids, mismatched UUIDs, alive vs. dead parents, expired tokens).

Regression:
- All existing tests pass.
- Existing `oc_stop` flow unchanged (Phase 4's `killProcessTree` is reused).
- `oc_session_resume` smoke test with `OPENCHROME_KILL_ON_EXIT=auto` (default).

## Acceptance Criteria

- [ ] After graceful agent exit (SIGINT/SIGTERM/stdin close), `ps aux` shows zero openchrome-launched Chrome processes within 5 seconds (unless session-resume token holds them).
- [ ] After ungraceful exit (SIGKILL), the next openchrome startup reaps the orphans within 10 seconds.
- [ ] User's daily-driver Chrome (no marker file) is never affected by reaping (verified by the explicit safety test in §Verification).
- [ ] On macOS, no `Google Chrome Helper` subprocesses survive (PGID kill cascades).
- [ ] PID files / markers do not accumulate (each session cleans up its own at exit; reaper sweeps stragglers).
- [ ] Repeated 10× session start/stop cycles show flat (not growing) RAM in `ps aux`.
- [ ] Attach mode (#659) does not write markers and never kills attached Chrome.
- [ ] Quiesced Chrome (#660) is not reaped while parent MCP is alive (parent-alive check verified by integration test).
- [ ] `OPENCHROME_KILL_ON_EXIT=never` opts out for session-resume use cases.
- [ ] Session-resume token format documented and exercised end-to-end.
- [ ] `oc_reap_orphans` tool implemented for manual sweep.
- [ ] macOS reaper uses `ps -ww` and refuses to kill on truncated/ambiguous cmdline output.

## Severity

**Highest of the five reported issues.** Recommend tackling first along with #657 (headless default flip).

## Files

- `src/transports/stdio.ts:80-91`
- `src/chrome/launcher.ts:605-642, 763-776`
- `src/utils/process-guardian.ts:76-77`
- `src/utils/pid-manager.ts:173-215`
- `src/mcp-server.ts:1756-1841`
- `src/tools/shutdown.ts:34-100`
- `src/index.ts:215, 252-263, 729-758`
- (new) `src/utils/process-tree.ts`
- (new) `src/chrome/ownership-marker.ts`
- (new) `src/tools/session-resume-token.ts`

shaun0927:
Closes #661.

## Summary

Eliminates Chrome / Helper / Renderer / GPU process accumulation across Claude Code and Codex sessions. Each MCP shutdown — graceful or stdin EOF — now reliably terminates the Chrome processes openchrome itself spawned. Next-startup reaping cleans up survivors of ungraceful exits (SIGKILL, OOM, terminal-close-without-handler).

## Why this matters

Per #661 root-cause analysis, the lifecycle gap was in the synchronous `process.exit(0)` path triggered by stdin EOF: the existing async cleanup never ran, and Chrome (spawned `detached: true` and `unref()`'d) outlived its parent MCP. Across a working day this accumulated several GB of orphaned Chrome processes, degrading the user's machine.

## Approach (from issue's phased plan)

| Phase | Status | Where |
|---|---|---|
| 1. Ownership marker file | ✅ | `src/chrome/ownership-marker.ts` (new) |
| 2. Sync best-effort kill on stdio EOF | ✅ | `src/utils/sync-shutdown.ts` (new) + `src/transports/stdio.ts` |
| 3. `process.on('exit')` safety net | ✅ | already present in `src/index.ts:229` (kept) |
| 4. Cross-platform `killProcessTree` | ✅ | already present in `src/utils/pid-manager.ts:105` (reused) |
| 5. Aggressive orphan reaper (4-way check) | ✅ | `src/utils/pid-manager.ts` `reapOrphanMarkers` + `classifyMarker` |
| 6. Attach-mode safety | ✅ | `launchMode` field on `ChromeInstance`; close() and reaper both skip `'attach'` |
| 7. Session-resume token | ✅ | `src/utils/session-resume-token.ts` (new) + `OPENCHROME_KILL_ON_EXIT` env |
| 8. Reconcile `detached: true` | ✅ | preserved (necessary for resume); now safe via marker scoping |

## Foundation for future PRs

- Adds `LaunchMode = 'isolated' \| 'attach'` and `instance.launchMode` field — directly reused by #660 (watchdog quiesce) and #659 (attach mode).
- The detect-existing-Chrome-on-port path is already marked `'attach'` here, so any Chrome openchrome did not spawn is structurally protected from being killed.

## Backwards compatibility

| Scenario | Old | New |
|---|---|---|
| `oc_stop` | Full async teardown | Same (reuses `killProcessTree`) |
| User exits Claude Code (graceful SIGTERM/SIGINT) | Existing `process.on('exit')` killed Chrome | Same + sync helper now also runs from stdio EOF path |
| User exits Claude Code (stdin EOF) | `process.exit(0)` ran before async cleanup → Chrome leaked | Sync SIGTERM→200ms→SIGKILL → marker cleanup → exit |
| User SIGKILLs Claude Code | Chrome leaked until reboot | Next openchrome startup reaps via marker |
| User has their own Chrome running | Untouched | Untouched (no marker → reaper ignores) |
| `oc_session_resume` workflow | Not formally defined | `OPENCHROME_KILL_ON_EXIT=auto` + token file preserves Chrome across MCP restart |

## Env vars added

- `OPENCHROME_KILL_ON_EXIT=auto\|always\|never` (default `auto`)
- `OPENCHROME_SESSION_RESUME_TTL_MIN` (default 30)

## Files

- new: `src/chrome/ownership-marker.ts`
- new: `src/utils/session-resume-token.ts`
- new: `src/utils/sync-shutdown.ts`
- modified: `src/chrome/launcher.ts` (LaunchMode field, marker write/remove, sync registration)
- modified: `src/transports/stdio.ts` (sync kill before process.exit)
- modified: `src/utils/pid-manager.ts` (reapOrphanMarkers + classifyMarker)
- new: `tests/chrome/ownership-marker.test.ts` (5 cases)
- new: `tests/utils/session-resume-token.test.ts` (5 cases)
- new: `tests/utils/sync-shutdown.test.ts` (6 cases)

## Test plan

- [x] `npm run build` clean.
- [x] 16 new unit cases pass.
- [x] Full suite: 3597 passed, 86 skipped, 1 pre-existing flaky timing test in `chrome-detector.test.ts` (passes 21/21 in isolation; unrelated to this PR — see #524 history).

## Verification (manual, post-merge)

```bash
npm run build

# 1. Graceful exit — sync handler kills Chrome.
node dist/index.js --auto-launch &
PID=\$!; sleep 3
kill -TERM \$PID; sleep 2
ps aux | grep -- '--user-data-dir.*openchrome' | grep -v grep | wc -l   # → 0

# 2. Ungraceful exit — orphan; reaped on next startup.
node dist/index.js --auto-launch &
PID=\$!; sleep 3
kill -KILL \$PID; sleep 2
node dist/index.js --auto-launch &
PID=\$!; sleep 5
ps aux | grep -- '--user-data-dir.*openchrome' | grep -v grep | wc -l   # → 1 (the new one only)

# 3. Daily-driver Chrome safety — no marker, untouched.
google-chrome --user-data-dir=/tmp/my-real-chrome &
node dist/index.js --auto-launch &
PID=\$!; sleep 3; kill -TERM \$PID; sleep 2
# Verify the user's chrome is still alive.

# 4. Session-resume preservation.
node dist/index.js --auto-launch &
PID=\$!; sleep 3
# (in agent) call oc_session_snapshot — writes resume token
kill -TERM \$PID; sleep 2
# Chrome should still be alive (token preserves it under OPENCHROME_KILL_ON_EXIT=auto)
```

### DIFF
--- a/src/chrome/launcher.ts
+++ b/src/chrome/launcher.ts
@@ -643,11 +643,26 @@ export class ChromeLauncher {
         pidFilePath: getChromePidFilePath(port),
         label: 'managed-chrome',
       });
+      // #661 Phase 1+2 (gemini high review): write the ownership marker AND
+      // register for synchronous shutdown immediately after spawn — before
+      // waitForDebugPort. Otherwise, if the parent dies during the debug-port
+      // wait window, the orphaned Chrome would not be in the registry and
+      // would leak.
+      if (userDataDir) {
+        writeMarker({ chromePid: chromeProcess.pid, userDataDir });
+      }
+      registerManagedChrome({ pid: chromeProcess.pid, userDataDir });
     }
 
     // Log Chrome process exit for immediate diagnostics
     chromeProcess.once('exit', (code, signal) => {
       console.error(`[ChromeLauncher] Chrome process exited (code: ${code}, signal: ${signal})`);
+      // Symmetric cleanup: unregister + remove marker so a failed-launch Chrome
+      // (e.g. waitForDebugPort timeout) does not leave stale state behind.
+      if (chromeProcess.pid) {
+        unregisterManagedChrome(chromeProcess.pid);
+        removeMarker({ chromePid: chromeProcess.pid, userDataDir });
+      }
       // Clear cached instance so next ensureChrome() knows Chrome is gone
       this.instance = null;
       // Clear pendingProcess if this was the one we were tracking
@@ -692,13 +707,8 @@ export class ChromeLauncher {
     }
     this.pendingProcess = null; // Success — no longer pending
 
-    // Write ownership marker (#661 Phase 1). Only for spawn-based 'isolated' launches —
-    // attach-mode (above) leaves the user's Chrome alone and writes no marker.
-    let ownershipMarker: string | null = null;
-    if (chromeProcess.pid && userDataDir) {
-      ownershipMarker = writeMarker({ chromePid: chromeProcess.pid, userDataDir });
-    }
-
+    // Marker + registration already happened immediately after spawn (above).
+    // Here we just record the instance state.
     this.instance = {
       wsEndpoint,
       httpEndpoint: `http://127.0.0.1:${port}`,
@@ -706,14 +716,11 @@ export class ChromeLauncher {
       userDataDir,
       profileType,
       launchMode: 'isolated',
-      ...(ownershipMarker ? { ownershipMarker } : {}),
     };
 
     // Persist Chrome PID to disk for orphan detection
     if (chromeProcess.pid) {
       writeChromePid(port, chromeProcess.pid);
-      // #661: register for synchronous shutdown.
-      registerManagedChrome({ pid: chromeProcess.pid, userDataDir });
     }
 
     this._intentionalStop = false;
--- a/src/chrome/launcher.ts
+++ b/src/chrome/launcher.ts
@@ -13,7 +13,16 @@ import { spawnProcessGuardian } from '../utils/process-guardian';
 import { DEFAULT_VIEWPORT, DEFAULT_CHROME_LAUNCH_TIMEOUT_MS, DEFAULT_RESTORE_LAST_SESSION } from '../config/defaults';
 import { ProfileManager } from './profile-manager';
 import type { ProfileType } from './profile-manager';
+import { writeMarker, removeMarker } from './ownership-marker';
+import { registerManagedChrome, unregisterManagedChrome } from '../utils/sync-shutdown';
 export type { ProfileType } from './profile-manager';
+
+/**
+ * Lifecycle ownership of a Chrome process (#661).
+ * - 'isolated': openchrome spawned this Chrome and owns its lifecycle. Eligible for sync-kill on exit.
+ * - 'attach' (#659, future): attached to a user-running Chrome. NEVER killed by openchrome.
+ */
+export type LaunchMode = 'isolated' | 'attach';
 
 export interface ChromeInstance {
   wsEndpoint: string;
@@ -21,6 +30,10 @@ export interface ChromeInstance {
   process?: ChildProcess;
   userDataDir?: string;
   profileType?: ProfileType;
+  /** Lifecycle ownership (#661). Defaults to 'isolated' for our spawn-based path. */
+  launchMode?: LaunchMode;
+  /** Random per-launch UUID written into the ownership marker file. */
+  ownershipMarker?: string;
 }
 
 export interface LaunchOptions {
@@ -372,6 +385,8 @@ export class ChromeLauncher {
         wsEndpoint: existingWs,
         httpEndpoint: `http://127.0.0.1:${port}`,
         ...(pendingProc && pendingProc.exitCode === null && { process: pendingProc }),
+        // Connected to a Chrome we did not spawn — never kill on exit (#661 Phase 6).
+        launchMode: 'attach',
       };
       // Attached to user-started Chrome — assume real profile
       this.profileState = { type: 'real', extensionsAvailable: true };
@@ -677,17 +692,28 @@ export class ChromeLauncher {
     }
     this.pendingProcess = null; // Success — no longer pending
 
+    // Write ownership marker (#661 Phase 1). Only for spawn-based 'isolated' launches —
+    // attach-mode (above) leaves the user's Chrome alone and writes no marker.
+    let ownershipMarker: string | null = null;
+    if (chromeProcess.pid && userDataDir) {
+      ownershipMarker = writeMarker({ chromePid: chromeProcess.pid, userDataDir });
+    }
+
     this.instance = {
       wsEndpoint,
       httpEndpoint: `http://127.0.0.1:${port}`,
       process: chromeProcess,
       userDataDir,
       profileType,
+      launchMode: 'isolated',
+      ...(ownershipMarker ? { ownershipMarker } : {}),
     };
 
     // Persist Chrome PID to disk for orphan detection
     if (chromeProcess.pid) {
       writeChromePid(port, chromeProcess.pid);
+      // #661: register for synchronous shutdown.
+      registerManagedChrome({ pid: chromeProcess.pid, userDataDir });
     }
 
     this._intentionalStop = false;
@@ -744,11 +770,18 @@ export class ChromeLauncher {
       try { this.pendingProcess.kill(); } catch { /* ignore */ }
       this.pendingProcess = null;
     }
+    // Attach mode (#659, #661 Phase 6): we did not spawn this Chrome — never kill it.
+    if (this.instance?.launchMode === 'attach') {
+      console.error('[ChromeLauncher] close(): attach-mode Chrome left alive (user-owned).');
+      this.instance = null;
+      return;
+    }
     if (this.instance?.process) {
       console.error('[ChromeLauncher] Closing Chrome...');
       const proc = this.instance.process;
       const userDataDir = this.instance.userDataDir;
       const profileType = this.currentProfileType;
+      const chromePidForMarker = proc.pid;
 
       if (process.platform === 'win32' && proc.pid) {
         try {
@@ -803,6 +836,12 @@ export class ChromeLauncher {
         } catch {
           // Ignore cleanup errors
         }
+      }
+
+      // Remove ownership marker (#661 Phase 1).
+      if (chromePidForMarker) {
+        removeMarker({ chromePid: chromePidForMarker, userDataDir });
+        unregisterManagedChrome(chromePidForMarker);
       }
     }
     // Remove Chrome PID file before clearing instance
--- a/src/chrome/ownership-marker.ts
+++ b/src/chrome/ownership-marker.ts
@@ -37,7 +37,13 @@ export interface OwnershipMarker {
   launchMode: 'isolated';
 }
 
-function readParentCommand(pid: number): string {
+/**
+ * Read a process's command-line. Linux only reads from /proc; on macOS/Windows
+ * we have no portable cheap way, so we fall back to the current process's argv —
+ * but only when the requested pid IS the current process. For any other pid
+ * we return '' rather than misleading data (gemini medium review on #667).
+ */
+export function readParentCommand(pid: number): string {
   if (process.platform === 'linux') {
     try {
       // /proc/<pid>/cmdline is NUL-separated; never truncated.
@@ -47,8 +53,11 @@ function readParentCommand(pid: number): string {
       return '';
     }
   }
-  // Best-effort across platforms; we use this as a hint, not a security check.
-  return process.argv.join(' ');
+  // Best-effort across non-Linux platforms; only valid for the current process.
+  if (pid === process.pid) {
+    return process.argv.join(' ');
+  }
+  return '';
 }
 
 function primaryMarkerPath(userDataDir: string): string {
@@ -139,6 +148,7 @@ export function readMarker(filePath: string): OwnershipMarker | null {
     if (
       typeof parsed.pid !== 'number' ||
       typeof parsed.ppid !== 'number' ||
+      typeof parsed.ppidCommand !== 'string' ||
       typeof parsed.userDataDir !== 'string' ||
       typeof parsed.marker !== 'string' ||
       parsed.launchMode !== 'isolated'
@@ -217,10 +227,11 @@ export function verifyChromePidIdentity(pid: number, expectedUserDataDir: string
     try {
       const raw = fs.readFileSync(`/proc/${pid}/cmdline`, 'utf8');
       const argv = raw.split('\0').filter((s) => s.length > 0);
-      return argv.some((arg) =>
-        arg === `--user-data-dir=${expectedUserDataDir}` ||
-        arg.startsWith(`--user-data-dir=${expectedUserDataDir}`),
-      );
+      const exact = `--user-data-dir=${expectedUserDataDir}`;
+      // Path separator matters: a bare startsWith allows /tmp/chrome to match
+      // /tmp/chrome-2 (gemini medium / codex P2 review on #667).
+      const sepPrefix = `--user-data-dir=${expectedUserDataDir}${path.sep}`;
+      return argv.some((arg) => arg === exact || arg.startsWith(sepPrefix));
     } catch {
       return false;
     }
@@ -235,7 +246,15 @@ export function verifyChromePidIdentity(pid: number, expectedUserDataDir: string
       });
       // ps -ww output is a single line for one PID.
       const cmdline = out.toString().trim();
-      if (!cmdline.includes(`--user-data-dir=${expectedUserDataDir}`)) {
+      const exact = `--user-data-dir=${expectedUserDataDir}`;
+      const sepPrefix = `--user-data-dir=${expectedUserDataDir}${path.sep}`;
+      // Same separator-aware match as the Linux path; reject prefix collisions
+      // like /tmp/chrome ↔ /tmp/chrome-2 (codex P2 review on #667).
+      const matches =
+        cmdline.includes(`${exact} `) ||
+        cmdline.endsWith(exact) ||
+        cmdline.includes(sepPrefix);
+      if (!matches) {
         // No match — could be PID reuse, or could be macOS truncation despite -ww.
         return false;
       }
--- a/src/chrome/ownership-marker.ts
+++ b/src/chrome/ownership-marker.ts
@@ -163,34 +163,76 @@ export function readMarker(filePath: string): OwnershipMarker | null {
 
 /**
  * Discover all known markers.
- * - Primary: walks `~/.openchrome/profiles/*` for `.openchrome-managed` files.
- * - Fallback: every `~/.openchrome/state/markers/*.json`.
- * Custom user-data-dirs that live elsewhere are NOT scanned (we only know
- * about ones we wrote into our own profile root).
+ *
+ * Scans (deduplicated):
+ *   - `~/.openchrome/profile/`              — singular persistent profile from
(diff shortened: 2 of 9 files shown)


shaun0927:
## Bot-review follow-up — merge ready ✅

Addressed in commit `812f679`:

- **gemini high** (`launcher.ts:716`): hoisted `writeMarker()` + `registerManagedChrome()` to fire immediately after `spawn()`, before `waitForDebugPort`, closing the orphan window if the parent exits during port binding. Symmetric cleanup in the `chromeProcess 'exit'` handler so a failed launch (port-wait timeout) doesn't leave stale state.
- **gemini medium / codex P2** (`ownership-marker.ts:222`, `:236`): replaced `startsWith('--user-data-dir=' + path)` with exact-match-or-`path.sep`-prefix logic on both Linux and macOS branches; prevents `/tmp/chrome` from falsely matching `/tmp/chrome-2` during PID-reuse identity verification.
- **gemini medium** (`ownership-marker.ts:51`): `readParentCommand()` now returns `process.argv` only when `pid === process.pid` on non-Linux platforms (instead of misleadingly returning the current process's argv for any pid). Function exported for future reaper use.
- **gemini medium** (`ownership-marker.ts:145`): `readMarker()` now validates `ppidCommand: string` so the reaper's PID-reuse defense always has the field it needs.

11/11 ownership-marker + 6/6 sync-shutdown unit tests pass; full suite 3598 passing locally. CI shows 1 unrelated pre-existing timing flake (`tests/cdp/tab-health-monitor.test.ts`, see PR #524 history).

Ready to merge.

shaun0927:
## Round-2 follow-up — codex re-review addressed ✅

After the round-1 fix (commit `812f679`), codex re-reviewed and surfaced two more P1 findings, both genuine:

Addressed in commit `72313ff`:

- **codex P1** (`ownership-marker.ts:174` "Scan the actual marker directory used by launched Chrome"): `listMarkers()` walked `~/.openchrome/profiles/*` (PLURAL) but `ProfileManager.resolveProfile()` writes the persistent profile at `~/.openchrome/profile/` (SINGULAR). The reaper therefore could not see any of the markers actually produced by real launches — orphan-reaping silently no-op'd. Fix: scan the singular path + sub-dirs, the plural path (forwards-compat), `~/.openchrome/headless-shell-profile/`, plus the fallback marker store. Deduplicate via a Set.

- **codex P1** (`pid-manager.ts:reapOrphanMarkers` "Honor session-resume tokens before reaping orphan markers"): the reaper never consulted `listAllTokens()`. With `OPENCHROME_KILL_ON_EXIT=auto`, a process intentionally skips sync-kill when an active token covers its Chrome — but the very next startup would still reap that Chrome via the marker path, breaking the resume contract. Fix: pre-load every alive token's `chromePid` into a protected Set and skip those markers; best-effort (token-store read errors don't stop reaping).

CI all-green; 16/16 marker / sync-shutdown / token unit tests pass; full suite clean.

Ready to merge.

PR #669, opened 2026-05-01
feat(#658): generic login-failure detection in fill_form

### LINKED ISSUE #658: Login flows silently fail — no detection of failed authentication
## Summary

Login attempts can complete with `success: true` even when the user is not actually authenticated. The codebase has **no login-failure detection**: it watches for redirects to known auth domains but never for *failed* auth states. Combined with multiple `catch {}` patterns in `navigate.ts`, `act.ts`, and `fill-form.ts`, a broken login is invisible to the agent calling the tool.

> Naming convention used in this issue: file paths use hyphens (`fill-form.ts`, `login-detector.ts`); MCP tool names exposed over JSON-RPC use underscores (`fill_form`, `act`). Both refer to the same code.

## Symptoms

A reproducible failure pattern:
1. Agent calls the `fill_form` tool with deliberately wrong credentials on any login page.
2. Tool response: `{ success: true, ... }`.
3. Page actually shows "Invalid email or password" or remains on the login form.
4. Subsequent tool calls fail mysteriously because the session isn't authenticated.

In headless mode this is amplified — there is no visual ground truth for the agent to fall back on.

### 1. No login-failure detection logic exists
- `src/smart-goto.ts:67-87` and `src/tools/navigate.ts:462-489` only detect **redirects to** known auth domains. They never detect:
  - HTTP 401/403/422 responses from the form-submit POST.
  - DOM error banners ("Invalid …", "incorrect …", "locked", "try again").
  - Login form re-display (a near-universal failure signal).
  - Hidden error messages revealed only after submit.

### 2. fill-form silently swallows submit failures
`src/tools/fill-form.ts:419`:

```ts
isError: errors.length > 0 && filledFields.length === 0
```

If **any** field was filled, `isError` is `false` regardless of submit-button errors. Errors go into a list but never bubble to the response.

### 3. "Non-fatal" catches in act/interact
- `src/tools/act.ts:330-332`, `:297-299` — `catch { /* Non-fatal — best effort */ }` returns success.
- `src/tools/act.ts:526-540` — only sets `failedAt` for `outcome === 'EXCEPTION' | 'ELEMENT_NOT_FOUND' | 'TIMEOUT'`. The act tool **does not currently emit `SILENT_CLICK`** at all; that classification is defined in `src/utils/ralph/outcome-classifier.ts:18` and used by Ralph engine, not by the act step loop. A failed login submit therefore appears as a non-error success.

### 4. Block-detection errors swallowed
`src/tools/navigate.ts:496, 553, 651, 686, 810` use:

```ts
detectBlockingPage(...).catch(e => { console.error(e); return null; })
```

Returning `null` means the navigation result lies about what happened.

### 5. Retry list excludes login failures
`src/tools/navigate.ts:24` — `RETRYABLE_BLOCK_TYPES = ['access-denied', 'bot-check', 'captcha']`. Login failures aren't in that list.

## Design Constraints

1. **Don't break successful logins**: the detector must have near-zero false-positive rate on real-world successes. False positives break agents in production.
2. **General, not site-specific**: no per-site config, no allowlists. Use signals that work across IdPs.
3. **Localization-aware**: cannot rely on English-only error text. Use HTTP status, URL, and DOM-structure signals first; text patterns are a last-ditch hint, not a primary signal.
4. **On-by-default with explicit opt-out**: detection runs by default (`loginCheck: 'auto'`). Existing tool consumers that intentionally probe failed logins (e.g., test harnesses) opt out via `loginCheck: 'off'`. The behavior change (turning some `success: true` into `success: false` on actually-failed logins) is the intended fix; the opt-out is the escape hatch.

### Phase 1 — Add a `LoginOutcome` detector (no behavior change yet)

Implement `detectLoginOutcome(page, formMeta): LoginOutcome` in `src/tools/login-detector.ts` (new file).

`LoginOutcome = 'success' | 'failed' | 'unknown'`

Decision matrix (highest-confidence signal wins; ties → `unknown`):

| Signal | Weight | Outcome |
|---|---|---|
| URL navigated away from login origin AND no error fragment | strong | `success` |
| Form-submit network response 200 + new Set-Cookie that wasn't there before | strong | `success` |
| Form-submit network response 4xx | strong | `failed` |
| Same login form still mounted on page after submit (form selector still resolves, password field still empty/full) | strong | `failed` |
| New element with `[role="alert"]`, `aria-live="assertive"`, or `[data-testid*="error"]` appeared with non-empty text | medium | `failed` |
| Body text contains language-agnostic markers (HTTP status text, structured error JSON) | medium | `failed` |
| (last resort) Body text matches multilingual error glossary (en, ko, ja, zh-CN, es, de, fr, pt) | weak | `failed` |
| None of the above | — | `unknown` |

The "still on login form" signal is the workhorse — it's site-agnostic and works in any locale.

### Phase 2 — Wire the detector

After every `fill_form` submit AND every `act` step whose target matches the submit-button heuristic, call the detector. The submit-button heuristic is precise:

```ts
function isSubmitButton(el: ElementHandle): boolean {
  // 1. native submit
  if (matches(el, 'input[type="submit"], button[type="submit"]')) return true;
  // 2. ARIA-role button or generic button inside a form that has a password field
  const form = el.evaluate(e => e.closest('form'));
  if (!form) return false;
  const hasPassword = form.evaluate(f => !!f.querySelector('input[type="password"]'));
  const isButtonish = matches(el, 'button, [role="button"]');
  return hasPassword && isButtonish;
}
```

Shadow DOM traversal is **not** required for this heuristic — login forms inside closed shadow roots are rare and out of scope; document the limitation.

Behavior after the detector runs:
- `success` → existing behavior, `success: true`.
- `failed` → tool response `success: false`, `errorReason: 'login_failed'`, with the strongest matching signal in the response so the agent can reason.
- `unknown` → existing behavior (don't regress).

### Phase 3 — Stop swallowing `detectBlockingPage` errors

Replace each `.catch(e => { console.error(e); return null; })` in `src/tools/navigate.ts` with a typed result `{ ok: false, reason: 'detector-error', error }` so callers can decide.

### Phase 4 — Add post-click DOM-change detection in act.ts

Because `act.ts` does not currently produce `SILENT_CLICK`, two approaches are valid; **pick option (a) for this issue**:

- **(a) chosen**: After a click step in `act.ts`, call `outcomeClassifier.classify(beforeSnapshot, afterSnapshot)` from `src/utils/ralph/outcome-classifier.ts`. If the classifier returns `SILENT_CLICK` AND the click target matches `isSubmitButton`, set `failedAt` on the step and emit `errorReason: 'silent_click_on_submit'`.
- (b) rejected: duplicate the DOM-diff logic inline in act.ts. Don't do this — keep one source of truth.

The change is localized to `src/tools/act.ts:526-540` (step error handling) plus a new import.

### Phase 5 — Optional: add `'login-failed'` to `RETRYABLE_BLOCK_TYPES`

Once detection is in place, optionally extend the retry path. **Default off** — retrying a login is dangerous (rate limits, account lockouts). Make it opt-in via a `retryLoginOnce: boolean` flag on the tool inputs.

## Backwards Compatibility

- Tools that currently return `success: true` for an actually-failed login will start returning `success: false`. **This is the intended fix**, but document it loudly in release notes as a behavior change.
- For callers that need the old behavior (e.g., they parse the page themselves), expose `loginCheck: 'auto' | 'off'` on `fill_form` and `act`. Default `'auto'`. Setting `'off'` restores pre-change behavior.
- No API additions to `navigate`; the `detectBlockingPage` typed-result change is internal.

## Edge Cases & Risks

1. **2FA / step-up auth**: the form is replaced with a 2FA prompt. The "still on login form" check would say `success` (form gone). Detector treats this as `success` of the password phase — correct.
2. **Magic-link / passwordless**: submit succeeds and shows "Check your email". URL didn't change but form is gone and a confirmation banner appeared. Detector returns `unknown` (no negative signal); we don't surface a false failure.
3. **SPA logins that submit via XHR with no navigation**: the URL doesn't change, but the form unmounts and the page transitions. Form-still-mounted check catches the failure case; the success case (form unmounts) yields `unknown` rather than `failed` — acceptable.
4. **CAPTCHA after submit**: separate signal; existing `bot-check` detector handles it.
5. **Localization**: primary signals are language-agnostic (HTTP status, form-still-mounted, ARIA roles). Text glossary is fallback only.
6. **Account lockout**: returns `failed` correctly; agent can choose to back off.
7. **Multi-step wizards (email → password → 2FA)**: the detector sees each step's submit independently. Success of step 1 (email accepted, advance to password page) reads as `success` (form changed); failure of step 1 (email rejected) reads as `failed` (form re-rendered).
8. **Closed shadow DOM logins**: out of scope; document the limitation.

## Verification Procedure

Add a fixture page server in `tests/fixtures/login-server/` (Express or static HTML) with these endpoints:
- `/login-success` — accepts any creds, redirects to `/dashboard` and sets cookie.
- `/login-fail-400` — returns 400 + same-page error banner.
- `/login-fail-same-page` — returns 200 but re-renders the form with an `[role="alert"]`.
- `/login-2fa` — accepts and shows 2FA challenge.
- `/login-magic-link` — accepts and shows "Check your email".
- `/login-spa-fail` — XHR submit, no nav, form re-renders with error.
- Translations: include `?lang=ko|ja|zh|es|de|fr|pt` query params on the failure pages with localized error text.

Test matrix:
- Each endpoint × `loginCheck: 'auto' | 'off'` → expected outcome documented.

Manual:
```bash
npm run build
node dist/index.js --auto-launch
# (in agent) call fill_form against each fixture URL with deliberately wrong creds; verify response shape.
```

Regression:
- Run full test suite. Pre-existing tests that passed wrong creds and asserted `success: true` need to be updated (intentional behavior change) — list each affected test in the PR.

## Acceptance Criteria

- [ ] `detectLoginOutcome()` exists with the documented decision matrix and unit tests (one per matrix row × happy/sad path).
- [ ] All six fixture endpoints produce the documented outcomes.
- [ ] At least 7 locales covered in the multilingual glossary fallback (en, ko, ja, zh-CN, es, de, fr, pt).
- [ ] `loginCheck: 'auto' | 'off'` parameter on `fill_form` and `act`.
- [ ] No `.catch(e => return null)` in `src/tools/navigate.ts`.
- [ ] Click step in `act.ts` invokes the ralph outcome classifier and treats `SILENT_CLICK` on submit-target as `failedAt`.
- [ ] `isSubmitButton` heuristic (Phase 2 code block) implemented and unit-tested.
- [ ] Smoke test: a real login on a public IdP (e.g., GitHub or Google staging) with correct creds still returns `success: true`.
- [ ] PR description lists each pre-existing test changed and the reason.

## Files

- `src/tools/navigate.ts:24, 462-489, 496, 553, 651, 686, 810`
- `src/tools/fill-form.ts:344-419`
- `src/tools/act.ts:297-299, 330-332, 526-540`
- `src/utils/ralph/outcome-classifier.ts:18` (read-only reference)
- `src/smart-goto.ts:67-117`
- `src/cdp/client.ts:1611-1625`
- (new) `src/tools/login-detector.ts`
- (new) `tests/fixtures/login-server/`

shaun0927:
Closes #658 (focused scope — see "Out of scope" below).

## Summary

Today, `fill_form` returns `success: true` even when the password was wrong, because nothing in the codebase looks for "you got bounced back" signals. This PR adds a small, site-agnostic detector and wires it into the post-submit path of `fill_form` so a failed login surfaces as `isError: true` with a clear reason.

## Decision matrix (intentionally conservative)

False positives are worse than false negatives — they break working flows. The detector only returns `'failed'` on a strong signal; ambiguous cases stay `'unknown'`.

| Signal | Outcome |
| --- | --- |
| Cross-origin navigation away from login origin | success |
| Same-origin path change AND form unmounted | success |
| Login form still mounted at the same URL | **failed** |
| Login form still mounted with ARIA-error banner | **failed** (banner in reason) |
| Form unmounted at same URL (SPA / 2FA / magic-link) | unknown — DO NOT regress |
| URL parse failure / detector exception | unknown |

The detector **only runs when the submitted form contained a password field**. Non-login submits keep their previous behavior unchanged.

## Changes

- new: `src/tools/login-detector.ts`
  - `detectionScript` — in-page DOM signals
  - `classifyLoginSignals(signals, input)` — pure decision matrix (unit-testable)
  - `detectLoginOutcome(page, input)` — high-level helper
  - Exports `LoginOutcome`, `LoginDetectResult`, `InPageSignals`
- modified: `src/tools/fill-form.ts`
  - new optional input parameter `loginCheck: 'auto' | 'off'` (default `'auto'`)
  - capture pre-submit URL/origin before the click
  - after settle, if the form had a password field, run the detector
  - emit `Login appears to have failed: …` line in tool text
  - flip `isError: true` when detector returns `'failed'`
  - attach `_meta.errorReason='login_failed'` so callers can branch
- new: `tests/tools/login-detector.test.ts` (11 cases)

## Out of scope (deferred to follow-up issues)

The full #658 spec listed five phases. This PR ships **Phase 1 + 2** (detector + fill_form integration). Other phases are deferred to keep the diff focused and reviewable:

- ⏭️ Phase 3 — `navigate.ts` `.catch(e => null)` audit. Wide blast radius; will be its own PR.
- ⏭️ Phase 4 — `act.ts` `SILENT_CLICK` reclassification on submit-button targets. Touches the step-loop; deserves its own review.
- ⏭️ Phase 1 (weak signals) — multilingual error glossary, HTTP 4xx / Set-Cookie network checks. Easy extensions of the matrix; defer until needed by a real failure mode.

The current Phase 1 + 2 already eliminates the original symptom (silent `success: true` on wrong password) for `fill_form`, which is the user-reported case.

## Backwards compatibility

| Scenario | Old | New |
|---|---|---|
| `fill_form` on a non-login form | unchanged | **unchanged** (no password field → detector skipped) |
| `fill_form` on login with correct creds and navigation | `success: true` | `success: true` (cross-origin → detector returns `success`) |
| `fill_form` on login with wrong creds, form re-renders | `success: true` (silent fail) | `isError: true`, `errorReason: 'login_failed'` |
| `loginCheck: 'off'` | n/a (new param) | restores pre-#658 behavior |
| Detector exception (e.g. detached frame mid-detection) | n/a | falls through as `'unknown'` — no regression |

## Test plan

- [x] `npm run build` clean.
- [x] 11 new `login-detector` cases pass.
- [x] Full suite: 3593 passed, 86 skipped, 0 failed.

## Manual verification (post-merge)

- A login flow with correct credentials on any public IdP → `success: true` (no regression).
- A login flow with deliberately wrong credentials → `isError: true`, response text starts `Login appears to have failed: …`.
- 2FA challenge after correct password → `unknown` (so the agent can proceed to the next step without false-failing).

### DIFF
--- a/src/tools/fill-form.ts
+++ b/src/tools/fill-form.ts
@@ -15,6 +15,7 @@ import { resolveElementsByAXTree, invalidateAXCache } from '../utils/ax-element-
 import { getTargetId } from '../utils/puppeteer-helpers';
 import { normalizeQuery } from '../utils/element-finder';
 import { humanType, humanMouseMove } from '../stealth/human-behavior';
+import { detectLoginOutcome, LoginDetectResult } from './login-detector';
 
 const definition: MCPToolDefinition = {
   name: 'fill_form',
@@ -49,6 +50,11 @@ const definition: MCPToolDefinition = {
         type: 'number',
         description: 'Poll interval in ms (50-2000). Default: 300',
       },
+      loginCheck: {
+        type: 'string',
+        enum: ['auto', 'off'],
+        description: 'After submit, run a generic login-failure detector that flips success → failure when the password form is still mounted. Default: "auto". Set "off" to restore pre-#658 behavior.',
+      },
     },
     required: ['tabId', 'fields'],
   },
@@ -67,6 +73,7 @@ const handler: ToolHandler = async (
   const clearFirst = args.clear_first !== false; // Default to true
   const waitForMs = args.waitForMs as number | undefined;
   const pollInterval = Math.min(Math.max((args.pollInterval as number) || 300, 50), 2000);
+  const loginCheck: 'auto' | 'off' = (args.loginCheck === 'off') ? 'off' : 'auto';
 
   const sessionManager = getSessionManager();
 
@@ -341,6 +348,7 @@ const handler: ToolHandler = async (
       }
 
       // Optional: Click submit button
+      let loginResult: LoginDetectResult | null = null;
       if (submit && filledFields.length > 0) {
         try {
           const submitLower = normalizeQuery(submit);
@@ -374,9 +382,32 @@ const handler: ToolHandler = async (
           }, submitLower), 10000, 'fill_form', context);
 
           if (submitButton) {
+            // #658: capture pre-submit URL/origin BEFORE the click for the login-outcome detector.
+            const preSubmitUrl = page.url();
+            let preSubmitOrigin = '';
+            try { preSubmitOrigin = new URL(preSubmitUrl).origin; } catch { preSubmitOrigin = ''; }
+
             await page.mouse.click(Math.round(submitButton.x), Math.round(submitButton.y));
             submitted = true;
             await new Promise(resolve => setTimeout(resolve, DEFAULT_FORM_SUBMIT_SETTLE_MS));
+
+            // #658: Detect generic login failure (form still mounted, error banner, no nav).
+            // Only runs when (a) the user opted in (loginCheck !== 'off'), (b) the form
+            // contained a password field. We check (b) by sniffing the page after submit;
+            // forms without a password field skip the detector entirely so non-login
+            // submissions are unaffected.
+            if (loginCheck === 'auto') {
+              try {
+                const hasPasswordField = await page.evaluate(
+                  () => document.querySelector('input[type="password"]') !== null,
+                );
+                if (hasPasswordField) {
+                  loginResult = await detectLoginOutcome(page as any, { preSubmitOrigin, preSubmitUrl });
+                }
+              } catch {
+                // Detector errors are non-fatal — leave loginResult null.
+              }
+            }
           } else {
             errors.push(`Could not find submit button matching "${submit}"`);
           }
@@ -385,7 +416,7 @@ const handler: ToolHandler = async (
           errors.push(`Failed to submit: ${e instanceof Error ? e.message : String(e)}`);
         }
       }
-      return { submitted };
+      return { submitted, loginResult };
     }, { settleMs: 200 });
 
     // Build compact result message
@@ -409,6 +440,21 @@ const handler: ToolHandler = async (
       resultParts.push(`Errors: ${errors.join('; ')}`);
     }
 
+    // #658: surface login-detector outcome.
+    const loginOutcome = formResult.loginResult;
+    if (loginOutcome && loginOutcome.outcome === 'failed') {
+      resultParts.push(`Login appears to have failed: ${loginOutcome.reason}`);
+    } else if (loginOutcome && loginOutcome.outcome === 'success') {
+      resultParts.push(`Login appears successful: ${loginOutcome.reason}`);
+    }
+
+    // Failure conditions (in priority order):
+    //   1. Submit was attempted, the page still shows the login form, and the
+    //      detector classified the outcome as 'failed' (#658).
+    //   2. Errors collected during fill AND no fields were filled.
+    const detectorFailedLogin = loginOutcome?.outcome === 'failed';
+    const isError = detectorFailedLogin || (errors.length > 0 && filledFields.length === 0);
+
     return {
       content: [
         {
@@ -416,8 +462,9 @@ const handler: ToolHandler = async (
           text: resultParts.join('\n') + (delta || ''),
         },
       ],
-      isError: errors.length > 0 && filledFields.length === 0,
-    };
+      isError,
+      ...(detectorFailedLogin ? { _meta: { errorReason: 'login_failed', loginCheckReason: loginOutcome.reason } } : {}),
+    } as MCPResult;
   } catch (error) {
     return {
       content: [
--- a/src/tools/fill-form.ts
+++ b/src/tools/fill-form.ts
@@ -365,6 +365,11 @@ const handler: ToolHandler = async (
           // Find submit button + report whether its enclosing <form> contains a
           // password field. The form-scoped check rules out unrelated password
           // inputs elsewhere on the page.
+          //
+          // codex P2 review on #669: a submit button can target a form via the
+          // `form` attribute while being OUTSIDE that form's DOM subtree. We
+          // therefore prefer `el.form` (HTMLFormElement-style API) when
+          // present and fall back to `el.closest('form')`.
           const submitButton = await withTimeout(page.evaluate((query: string): { x: number; y: number; formHasPassword: boolean } | null => {
             const queryLower = query.toLowerCase();
             const selectors = [
@@ -384,7 +389,11 @@ const handler: ToolHandler = async (
                 if (text.includes(queryLower) || queryLower.includes(text.trim())) {
                   const rect = el.getBoundingClientRect();
                   if (rect.width > 0 && rect.height > 0) {
-                    const form = el.closest('form');
+                    // Prefer the HTMLButtonElement / HTMLInputElement `.form`
+                    // accessor — handles `<button form="loginForm">` cases that
+                    // `closest('form')` misses.
+                    let form: HTMLFormElement | null = (el as HTMLButtonElement | HTMLInputElement).form ?? null;
+                    if (!form) form = el.closest('form');
                     const formHasPassword = form
                       ? form.querySelector('input[type="password"]') !== null
                       : false;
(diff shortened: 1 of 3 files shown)


shaun0927:
## Bot-review follow-up — merge ready ✅

Addressed in commit `47059ca`:

- **gemini medium + codex P1** (`fill-form.ts:405`–`:410`): replaced the post-submit `document.querySelector('input[type="password"]')` global check with a **pre-submit, form-scoped** check: the submit-button finder now also reports whether its enclosing `<form>` contains a password field, captured before the click. This:
  - lets the success path fire even after the page navigates away (gemini medium);
  - prevents the detector from running on non-login forms whose page happens to contain an unrelated persistent password input, e.g. account-settings change-password forms (codex P1).
- **qodo Action#1** (`fill-form.ts:457` "Submit errors don't set isError"): track `submitErrored` separately from fill errors and include it in `isError`. A failed submit now correctly surfaces as an error even when some fields filled.
- **qodo Action#2** (`fill-form.ts:467` "errorReason not top-level"): `errorReason` is now a top-level field on the result (`'login_failed'` or `'submit_failed'`) so callers can branch without parsing text or reaching into `_meta`. `_meta` retained for back-compat.

qodo Action#3 (act-tool symmetric `loginCheck`) is acknowledged out of scope; PR description already lists Phase 4 (`act.ts` `SILENT_CLICK` reclassification) as a follow-up — the symmetric `loginCheck` lands together with that work.

11/11 detector unit tests pass; full suite 3592 passing locally (1 pre-existing port-conflict flake on `tests/watchdog/event-loop-monitor.test.ts`, passes 17/17 in isolation). CI all-green on this PR.

Ready to merge.

shaun0927:
## Round-2 follow-up — codex re-review addressed ✅

After the round-1 fix (commit `47059ca`), codex re-reviewed and surfaced two more issues, both real:

Addressed in commit `f9a7a44`:

- **codex P1** (`fill-form.ts:420` "Delay login-outcome classification until submit settles"): the detector ran exactly once after the 100ms `DEFAULT_FORM_SUBMIT_SETTLE_MS`, which is too tight — successful logins routinely sit on the form for >100ms while the auth POST round-trips, producing false `'failed'` on slow networks. Fix: poll up to 3 seconds, return EARLY on the first definitive `'success'` (cross-origin nav). `'unknown'` also short-circuits. Only `'failed'` waits the full window so a delayed redirect can flip the result. Worst case: 3s added to genuinely-failed login paths; happy path stays ~100ms.

- **codex P2** (`fill-form.ts:390` "Detect login forms linked via submit button `form` attribute"): `el.closest('form')` missed the HTML pattern where a `<button form="loginForm">` sits OUTSIDE the form's DOM subtree. In that layout `formHasPassword` was always false, so the detector never ran. Fix: prefer `el.form` (HTMLButtonElement / HTMLInputElement accessor that honors the `form` attribute), fall back to `closest('form')` for non-button roles.

11/11 detector tests pass; CI all-green.

Ready to merge.

PR #760, opened 2026-05-08
feat(perception): voting provider HTTP wrappers (anthropic + openai)

shaun0927:
## Summary

Concrete `Voter` HTTP wrappers for the multi-model voting framework introduced in **#759** (M3 PR-19). Adds two providers (Anthropic, OpenAI) behind the same interface so operators can configure two real models out of the box. Stacked on **#759** (closes M3 chain).

- `src/perception/providers/anthropic.ts` — `createAnthropicVoter({ model, apiKey, baseUrl? })` → `Voter`. Sends a structured prompt to the Messages API; expects `{ vote: "proceed"|"abstain", reason }` JSON; falls back to `abstain` on parse failure (which the voting layer treats as dissent → escalate)
- `src/perception/providers/openai.ts` — `createOpenAIVoter({ model, apiKey, baseUrl? })` with the same shape over the OpenAI Chat Completions API
- Both providers use `fetch` (no SDK dependency added). API key is read at voter-construction time; the request never logs the key
- `tests/perception/providers/*.test.ts` — request shape, response parsing, malformed-response handling, network-error handling (treated as `abstain`)

## Why no SDK dependency

#699/#700 epic posture: thin runtime, minimal supply chain. `fetch` + a 50-line wrapper is auditable and avoids dragging an entire SDK into the build for two endpoints. If we later need streaming or richer features we can revisit per provider.

## Why default to `abstain` on any failure

Aligns with the conservative posture of #759: any uncertainty by a `critical: true` voter resolves to escalation, never silent execution.

## Validation

- `npm run build` clean
- `npm test -- tests/perception` — all green
- No env-var requirement at module load (constructors take the key explicitly so unconfigured deployments do not crash)

## Test plan

- [ ] Reviewer pulls and runs `npm install && npm run build && npm test -- tests/perception`
- [ ] Reviewer confirms tests do NOT make real network calls (mock `fetch` is used)
- [ ] Reviewer confirms API key never appears in logged request representation (`request.toString()` etc.)
- [ ] Reviewer confirms the prompt schema is documented in the provider module so swapping models doesn't silently break vote shape

Refs: #700 (epic), #711 (sub-issue). Stacked on #759.

🤖 Split from `feat/m1-pr1-trace-foundation`. Original commit: `f4b88ec`.

PR #756, opened 2026-05-08
feat(contracts): beforeIrreversibleAction hook for critical contracts

shaun0927:
## Summary

Adds `beforeIrreversibleAction` hook for `critical: true` contracts. The runtime calls registered hook(s) immediately before dispatching any action that would commit money / change account state / send a message. Hooks can **veto, escalate to handoff, or pass through**. This is the integration seam that #711 (multi-model voting, Epic #700) plugs into. Stacked on **#755**.

- `src/contracts/runtime.ts`
  - New optional `args.hooks.beforeIrreversibleAction?: (ctx) => Promise<HookVerdict>`
  - Called only when `contract.critical === true` AND the next action is in the `IRREVERSIBLE_ACTION_KINDS` set (`fill_form` with `submit`, `act` with intent containing "purchase|checkout|confirm|send|delete|pay", explicit `irreversible: true` flag)
  - `HookVerdict` ∈ `proceed | veto({reason}) | escalate({deepLink, reason})`
- `tests/contracts/before-irreversible.test.ts` — every verdict path, hook timeout (default 5s, configurable), multiple hooks composition (all-must-proceed semantics)

## Why a hook (not a hard-coded rule)

#711's voting logic, perception cross-check, and any future policy engine all need to gate the same set of irreversible actions. A hook seam keeps the runtime's responsibilities narrow and lets policy ship as a separate, swappable concern.

## What is deferred

- The actual #711 voting policy that consumes this hook → M3 PR (separate)
- A registry of irreversible action kinds beyond the conservative built-in set → out of scope; callers can use `irreversible: true` on the action

## Validation

- `npm run build` clean
- `npm test -- tests/contracts` — all green
- Pure runtime change: zero behavior change for non-`critical` contracts (hook is never called)

## Test plan

- [ ] Reviewer pulls and runs `npm install && npm run build && npm test -- tests/contracts`
- [ ] Reviewer confirms hook is **not** called for `critical: false` contracts (zero cost on the hot path)
- [ ] Reviewer confirms hook timeout is enforced and treated as `escalate` (fail-safe, not fail-open)
- [ ] Reviewer confirms multi-hook composition: any `veto` short-circuits; any `escalate` (in absence of veto) wins over `proceed`

Refs: #699 (epic) and #700 epic (#711). Stacked on #755.

🤖 Split from `feat/m1-pr1-trace-foundation`. Original commit: `929132a`.

PR #749, opened 2026-05-08
feat(contracts): runtime core with verdict taxonomy + retry (M2 PR-11)

shaun0927:
## Summary

Wraps a skill function with pre/post-condition enforcement. **Always settles (never throws)**. Each call emits exactly one `TransactionRecord` through the audit emitter — wired through `logAuditEntry` by default, swappable via `args.audit` for tests. Stacked on **#748**.

- `src/contracts/runtime.ts`
  - `runUnderContract(contract, skill, args)` — single entry point
  - Verdict taxonomy per #706 v2: `success | precondition_failed | postcondition_failed | budget_exceeded | retry_exhausted | execution_threw | escalated`
  - Bounded retry per `contract.on_fail.retry` with idempotency-cache hook (PR-12) and budget enforcement (per `contract.budget.tokens` + `contract.budget.wall_ms`)
  - `TransactionRecord` carries: contract id, verdict, latency, token cost, retry count, the failing assertion id (if any), and a `traceSessionId` link to the recorder (#735+#744)
- `tests/contracts/runtime.test.ts` — every verdict path, retry-on-postcondition-failure, budget exhaustion mid-execution, audit emission shape, error isolation

## "Always settles" guarantee

The runtime catches everything from the wrapped skill function and converts it to a verdict. This is the safety property that makes contracts usable for paid actions: the caller can rely on receiving a `TransactionRecord` and never has to reason about partial failure.

## What is deferred

- Evidence bundle on failure (#707) — PR-13
- Idempotency cache implementation — PR-12 (this PR exposes the hook only)
- Headed-handoff escalation — PR-14/PR-15
- `screenshot_class` post-assertion → blocked on PR-10 pHash

## Validation

- `npm run build` clean
- `npm test -- tests/contracts` — all green
- Pure runtime: zero browser/network coupling. Skill function is an injected callable

## Test plan

- [ ] Reviewer pulls and runs `npm install && npm run build && npm test -- tests/contracts`
- [ ] Reviewer confirms `runUnderContract` never re-throws — every test path returns a `TransactionRecord`
- [ ] Reviewer confirms budget is checked **between retries** as well as during execution (no infinite-retry budget bypass)
- [ ] Reviewer confirms the audit emitter receives exactly one event per call (no double emission on retry)

Refs: #699 (epic), #706 (sub-issue). Stacked on #748.

🤖 Split from `feat/m1-pr1-trace-foundation`. Original commit: `5512f16`.

PR #738, opened 2026-05-08
feat(skill): per-domain skill graph SQLite storage (M1 PR-5)

shaun0927:
## Summary

Persistent backing store for #702's skill graph. **One SQLite database per domain** at `<rootDir>/<domain>.db` so concurrent activity on different domains is fully independent (only same-domain writes serialise via WAL + `BEGIN IMMEDIATE`). Stacked on **#737** (PR-4 hashing).

### Schema (v1) — pinned per #702 v2

To avoid a #703 follow-up migration, the schema includes `to_state_distribution` from day one:

```sql
nodes(state_hash PK, evidence_blob, thumbnail_path, last_seen_at, visit_count);
edges(from_state, action_kind, action_args_norm,
      to_state_distribution TEXT NOT NULL DEFAULT '[]',  -- JSON array of {to_state, count}
      success_count, fail_count, last_failed_at,
      PRIMARY KEY(from_state, action_kind, action_args_norm),
      FOREIGN KEY(from_state) REFERENCES nodes(state_hash));
applied_migrations(version PK, applied_at);
```

PK on edges is `(from_state, action_kind, action_args_norm)` — the distribution lives in a single row keyed that way, matching the v2 spec change so #703's executor can `UPDATE … SET to_state_distribution = …` without ever ALTERing.

### Files

- `src/skill/storage.ts` — `SkillGraphStorage` with `recordEdge`, `recordSuccess`, `recordFailure`, `getNode`, `topEdges`, `recentFailures`. WAL mode + `synchronous=NORMAL`. Per-domain DB pool keeps prepared statements warm.
- `src/skill/index.ts` — exports the storage symbol
- `tests/skill/storage.test.ts` — schema, migrations idempotence, edge upsert + distribution merging, success/fail counters, concurrent same-domain writes serialise correctly, cross-domain writes proceed independently

## Why per-domain DB (not single file)

#702 v2 lock-model note: "One database file per domain ⇒ writers on the same domain serialise via SQLite WAL mode and `BEGIN IMMEDIATE`. Cross-domain writes are independent." This avoids a single global lock that would serialise unrelated browsing.

## Validation

- `npm run build` clean
- `npm test -- tests/skill` — all green (no skill tests broken by storage addition)
- No coupling to recorder, contracts, or perception modules — pure storage layer

## Test plan

- [ ] Reviewer pulls and runs `npm install && npm run build && npm test -- tests/skill`
[...]

PR #761, opened 2026-05-08
feat(skill-memory): verified skill extractor (M4 PR-20)

shaun0927:
## Summary

Verified-skill extractor — turns successful, contract-verified runs into reusable, named skills. Skills are extracted **only from runs that the contract runtime (#749 + #756) marked as `success`** and only after a configurable confidence threshold. Stacked on **#760** (closes M3 chain).

- `src/skill-memory/extractor.ts`
  - `extract({ traceSessionId, transactionRecord }) → ExtractedSkill | null`
  - Walks the trace events, abstracts URL trail (uses #738 URL normalizer), groups consecutive `act/click/fill_form` calls into a step list, attaches the contract id and verdict
  - **Returns `null` on any failure heuristic** (verdict ≠ `success`, missing trace events, no contract attached) — never extracts from a doubtful run
- `src/skill-memory/types.ts` — `ExtractedSkill` shape
- `src/skill-memory/index.ts` — barrel
- `tests/skill-memory/extractor.test.ts` — happy path, refusal on failed verdict, refusal on missing trace, abstraction stability across two runs that differ only in tracking params

## Why "verified only"

#699 epic posture: never trust a self-reported success. The contract layer is the only signal we trust to mean "this actually worked." Skills extracted from unverified runs would poison the recall path (#761 / next PR).

## Validation

- `npm run build` clean
- `npm test -- tests/skill-memory` — all green
- Pure transform: input is `(traceSessionId, transactionRecord)`, output is a JSON skill or `null`

## Test plan

- [ ] Reviewer pulls and runs `npm install && npm run build && npm test -- tests/skill-memory`
- [ ] Reviewer confirms extractor returns `null` for any non-`success` verdict (no escape hatch)
- [ ] Reviewer confirms abstraction collapses two runs that differ only in `?utm_source=…` to the same step list
- [ ] Reviewer confirms extractor never reads/writes anything outside the trace storage layer (#735) and the new skill-memory storage (#761 next slice)

Refs: skill memory M4 PR-20. Stacked on #760.

🤖 Split from `feat/m1-pr1-trace-foundation`. Original commit: `fbf82ad`.

### DIFF
--- a/src/skill-memory/extractor.ts
+++ b/src/skill-memory/extractor.ts
@@ -160,7 +160,46 @@ export function recordSuccessfulRun(
   const windowStartMs = t - ROLLING_WINDOW_MS;
 
   const existing = fs.existsSync(filePath) ? parseSkillMd(fs.readFileSync(filePath, 'utf8')) : null;
-  const existingSidecar = readJson<SkillSidecar>(sidecarPath);
+  let existingSidecar = readJson<SkillSidecar>(sidecarPath);
+  // Markdown exists but sidecar is missing or unreadable. Falling
+  // through to the "create new" path would silently reset
+  // verified_runs to 1 and discard the promotion state recorded in
+  // the markdown, so reconstruct a minimal sidecar from the
+  // frontmatter instead. This preserves the count across transient
+  // sidecar IO issues; the rolling window collapses to a single
+  // synthetic entry timestamped at the previous `last_verified_at`.
+  if (existing && !existingSidecar) {
+    const priorTs = Date.parse(existing.frontmatter.last_verified_at);
+    const priorMs = Number.isFinite(priorTs) ? priorTs : t;
+    const priorRuns = Math.max(0, existing.frontmatter.verified_runs);
+    // Seed one synthetic recent entry per prior verified run so the
+    // success-count recomputation in the merge path lands on the same
+    // verified_runs total, which preserves promoted status across a
+    // missing sidecar. Capped at SKILL_RUN_LOG_MAX-1 to leave room for
+    // the new entry the merge path appends. The exact timestamps are
+    // unknown — we anchor at last_verified_at so the rolling-window
+    // eventually drops them naturally.
+    const seedCount = Math.min(priorRuns, SKILL_RUN_LOG_MAX - 1);
+    const recent: SkillSidecar['runs']['recent'] = [];
+    for (let i = 0; i < seedCount; i++) {
+      recent.push({
+        txn_id: existing.frontmatter.contract_ref,
+        ok: true,
+        ts: priorMs,
+      });
+    }
+    existingSidecar = {
+      schema_version: SKILL_SCHEMA_VERSION,
+      skill_id: skillId,
+      graph_node_anchor: existing.frontmatter.graph_node_anchor,
+      contract_id: inputs.contract_id,
+      runs: {
+        count: priorRuns,
+        window_start: isoUtc(windowStartMs),
+        recent,
+      },
+    };
+  }
 
   let frontmatter: SkillFrontmatter;
   let sidecar: SkillSidecar;
--- a/src/skill-memory/extractor.ts
+++ b/src/skill-memory/extractor.ts
@@ -178,6 +178,65 @@ function writeAtomic(target: string, body: string): void {
 }
 
 /**
+ * Resolve the sidecar to merge against for an update.
+ *
+ * - No prior markdown on disk → returns null (fresh-creation path).
+ * - Markdown present, sidecar valid → returns the parsed sidecar.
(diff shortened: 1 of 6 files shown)


shaun0927:
## Status note for reviewers — Codex bot re-flag cycle

This PR has gone through 6 rounds of Codex bot review. Each round, the bot has surfaced a mix of **real** issues (all addressed) and **stale re-flags** of patterns already fixed via different surface forms.

### Real issues addressed (cumulatively)
- P1: `parseSimpleYaml` no longer eagerly coerces digit-only / boolean-like strings; type coercion is deferred to validation time so string-typed fields (`contract_ref`, `graph_node_anchor`, `name: "true"`, etc.) round-trip cleanly (commits d4d0c40, 57163a7).
- P2: Sidecar reconstruction from frontmatter when sidecar is missing — preserves `verified_runs` + promotion state across transient IO issues (commit d4d0c40, refactored into `loadPriorSidecar` in d1a422c).
- P1: `isValidSidecar` shape guard — malformed sidecar (`{}`, older schemas, `null` recent entries) routes into the rebuild branch instead of crashing (commits bbc7a5a, 4b90017).
- P2: `writeAtomic` uses per-call unique temp paths (`{target}.{pid}.{rand}.tmp`) so concurrent writers cannot race (commit bbc7a5a).
- P1: `assertSafeDomain` blocks path traversal in both `recordSuccessfulRun` and `listSkillsForDomain` (commit 252293a).
- P1: `setNested` blocks `__proto__` / `constructor` / `prototype` segments and uses null-prototype objects (commit 252293a).
- P2: `promotionThreshold` is now bounded to `[1, SKILL_RUN_LOG_MAX]` so it cannot silently saturate (commit 4b90017).

### Stale re-flags (already covered, retained for context)
- "Validate domain before joining with skill root" / "Reject path traversal in domain reads" — both call sites have `assertSafeDomain(domain)` BEFORE the `path.join`. Tests: 7 parameterized cases (`../escape`, `..`, `amazon.com/../etc`, `amazon.com/sub`, `amazon.com\\sub`, `has spaces`, `""`) plus `listSkillsForDomain rejects malformed domain`.
- "Refuse update when sidecar is missing" / "Guard sidecar schema before reading runs.recent" — both flow through `loadPriorSidecar`, which returns the parsed sidecar when valid, rebuilds from frontmatter when missing/invalid, and validates each entry shape. Tests: `structurally malformed sidecar is treated as missing`, `sidecar with malformed recent entries is treated as missing`, `preserves verified_runs when sidecar disappears between runs`, `preserves promoted status when sidecar is missing`.

### Verification
- `npm run build` — clean
- `npx jest tests/skill-memory/` — 53/53 green (17 new tests added across the rounds)
- `mergeStateStatus: CLEAN`

shaun0927 (commented):
Addressed current-head Codex P1 findings for PR #761.

Findings addressed:
- Preserved existing SKILL.md body content when a successful rerun omits `inputs.body`, so distilled/curated bodies are not replaced by the placeholder.
- Serialized the per-skill read/merge/write persistence path with an atomic lock directory, preventing concurrent writer processes from dropping successful-run increments.

Validation:
- `npm test -- tests/skill-memory/extractor.test.ts`
- `git diff --check`
- `npm run build`

New head: `75190e76fb588054f9bfa05490d3a57f5074b4c8`

shaun0927:
Codex P1 remediation evidence for current head 20718033eb6c61df1d6c625522c237859a023aff:

- Fixed real serializer issue in commit 20718033eb6c61df1d6c625522c237859a023aff: `escapeStr` now quotes CR/LF-bearing frontmatter values, with a regression test proving `intent: "step 1\nstep 2"` round-trips as one value.
- The remaining current-head bot P1 inline comments are stale against this head:
  - domain traversal: `assertSafeDomain` rejects empty, long, separators, `.`/`..`, traversal segments, and forbidden characters before `path.join`; `listSkillsForDomain` uses the same guard.
  - malformed recent entries: `isValidSidecar` validates each `runs.recent` entry has finite numeric `ts`, boolean `ok`, and string `txn_id`; malformed sidecars rebuild from frontmatter instead of throwing.
  - body preservation: update path uses `inputs.body ?? existing?.body ?? placeholder`, so re-runs without a body preserve curated SKILL.md content.
  - serialized updates: `recordSuccessfulRun` wraps read/merge/write in `withSkillLock`, uses a per-skill lock directory, and atomic unique temp writes.

Validation run after the fix:
- `npx jest tests/skill-memory/extractor.test.ts tests/skill-memory/skill-md.test.ts --runInBand` -> 56 passed
- `git diff --check` -> passed
- `npm run build` -> passed

shaun0927:
Replaced by PR #810 (merged to develop).

PR #762, opened 2026-05-08
feat(skill-memory): skill recall + frozen-snapshot store (M4 PR-21)

shaun0927:
## Summary

Skill recall + frozen-snapshot store. Lets a future task ask "have I done this on this site before?" and get back a list of verified skills ranked by recency × success count, each pinned to a frozen DOM snapshot so the recall result is reproducible. Stacked on **#761**.

- `src/skill-memory/store.ts` — SQLite at `~/.openchrome/skill-memory/<domain>.db`
  - `skills(skill_id PK, domain, name, steps_json, contract_id, success_count, last_used_at, frozen_snapshot_path)`
  - `recall({ domain, currentStateHash?, currentUrl? }) → SkillRecallResult[]`
  - WAL mode; per-domain DB matches the #738 lock-model convention
- `src/skill-memory/frozen-snapshot.ts` — write-once DOM snapshots (`.json.gz` under `~/.openchrome/skill-memory/snapshots/`); referenced by `skills.frozen_snapshot_path`
- `tests/skill-memory/store.test.ts` + `tests/skill-memory/frozen-snapshot.test.ts`

## Why a separate DB (not reuse the skill graph DB from #738)

Different read patterns: the skill graph is keyed on `state_hash` for executor resume; skill memory is keyed on `(domain, name)` for human-discoverable recall. Mixing the two would mean every recall query joins against a much larger graph table. Per-DB isolation also means a skill-memory bug cannot corrupt the executor's resume index.

## Why frozen snapshots

Per #702 v2 capture procedure (#740): a recall result must be reviewable. The frozen snapshot lets an operator answer "what did the page look like when this skill last worked?" without spinning a browser.

## Validation

- `npm run build` clean
- `npm test -- tests/skill-memory` — all green
- Storage layer is independent: no behavior change for any other module

## Test plan

- [ ] Reviewer pulls and runs `npm install && npm run build && npm test -- tests/skill-memory`
- [ ] Reviewer confirms WAL mode set on every connection
- [ ] Reviewer confirms recall ordering is stable (recency × success count, deterministic tie-break)
- [ ] Reviewer confirms frozen snapshot is gzipped and write-once (no in-place edits)

Refs: skill memory M4 PR-21. Stacked on #761.

🤖 Split from `feat/m1-pr1-trace-foundation`. Original commit: `249073e`.

PR #759, opened 2026-05-08
feat(perception): multi-model voting + args equivalence (M3 PR-19)

shaun0927:
## Summary

Multi-model voting for `critical: true` contracts. When two configured perception models disagree on the next action, the runtime escalates rather than executes the irreversible action. Plugs into the `beforeIrreversibleAction` hook (#756). Stacked on **#758**.

- `src/perception/voting.ts`
  - `Voter` interface (`name`, `decide(snapshot, candidateAction) → VoterDecision`)
  - `vote(voters, snapshot, candidateAction) → VoteOutcome` — runs voters in parallel with bounded concurrency, deadlines, and one retry per voter
  - `argsEquivalence(a, b)` — canonical comparator for action args (whitespace, key order, semantic-equivalent selectors). Two voters that agree on the click target but spell its selector differently still count as agreeing
- `src/perception/voting-hook.ts`
  - Adapter that wires `vote(...)` into the runtime's `beforeIrreversibleAction` hook signature
  - `proceed` requires unanimous agreement; any `disagreement` → `escalate({reason:"voter-disagreement", deepLink})`; any voter timeout → `escalate({reason:"voter-timeout"})`
- `tests/perception/voting.test.ts` — unanimity, single dissent escalates, timeout escalates, args-equivalence tolerance, voter throw treated as dissent

## Why "any disagreement → escalate" (not majority)

Per #711 v2: "When two configured models disagree on the next action of a `critical: true` contract, the transaction escalates rather than executes." Majority-vote at N=2 is undefined; majority at N=3 introduces a tie-breaker policy decision that has no obvious right answer. Conservative escalation matches the security posture (irreversible actions get human review on doubt).

## What is deferred

- Concrete model HTTP wrappers (Anthropic, OpenAI) → next commit (`feat(perception): voting provider HTTP wrappers`)
- A registry of named voter configurations → out of scope; users wire voters explicitly today

## Validation

- `npm run build` clean
- `npm test -- tests/perception` — all green
- Pure orchestration: voters are injected, no network code in this PR

## Test plan

- [ ] Reviewer pulls and runs `npm install && npm run build && npm test -- tests/perception`
- [ ] Reviewer confirms `argsEquivalence` is order-insensitive and whitespace-insensitive
- [ ] Reviewer confirms voter timeout is bounded (a hung model cannot block escalation indefinitely)
- [ ] Reviewer confirms a thrown error in any voter is treated identically to a `disagreement` (fail-safe)

[...]

PR #763, opened 2026-05-08
feat(skill-memory): curator (Pass 1 + Pass 3) + PID lock (M4 PR-22)

shaun0927:
## Summary

Curator passes 1 + 3 with a PID-file lock. The curator periodically grooms the skill-memory store: **Pass 1 prunes** (drops skills past TTL or below confidence floor), **Pass 3 promotes** (recomputes recall ranking weights from accumulated `success_count` / `last_used_at`). A PID-file lock at `~/.openchrome/skill-memory/curator.lock` ensures only one curator runs against a domain even if the operator launches `oc serve` twice. Stacked on **#762**.

- `src/skill-memory/curator/lock.ts` — atomic PID-file lock with stale detection (process not alive → take over)
- `src/skill-memory/curator/prune.ts` — Pass 1
- `src/skill-memory/curator/promote.ts` — Pass 3
- `src/skill-memory/curator/runner.ts` — schedules passes; idempotent (a second invocation while one is running is a no-op via the lock)
- `tests/skill-memory/curator/*.test.ts`

## Why a PID lock (not flock)

Cross-platform: `flock(2)` is POSIX-only and openchrome targets macOS / Linux / Windows. PID + atomic rename works everywhere. Stale-detection covers crashed processes.

## What is deferred

- Pass 2 (sibling skill merge) → PR-23 (next slice). Uses an LLM merge requester so it ships independently
- A scheduling integration with the running MCP server → out of scope for this PR; the curator is invoked via the CLI today

## Validation

- `npm run build` clean
- `npm test -- tests/skill-memory` — all green
- Idempotence test: launching the runner twice in a row never double-prunes

## Test plan

- [ ] Reviewer pulls and runs `npm install && npm run build && npm test -- tests/skill-memory`
- [ ] Reviewer confirms PID lock survives stale processes (kill -9 a holder; another runner takes over within the documented stale window)
- [ ] Reviewer confirms Pass 1 never deletes a skill that was used within `OPENCHROME_SKILL_MEM_FRESH_DAYS` even if confidence is low (recency floor)
- [ ] Reviewer confirms Pass 3 ranking is deterministic for fixed `(success_count, last_used_at)`

Refs: skill memory M4 PR-22. Stacked on #762.

🤖 Split from `feat/m1-pr1-trace-foundation`. Original commit: `b09d4f3`.

### DIFF
--- a/src/skill-memory/curator-lock.ts
+++ b/src/skill-memory/curator-lock.ts
@@ -86,10 +86,7 @@ export class CuratorLock {
     }
     // Lock exists — decide whether to reclaim.
     if (this.shouldReclaim()) {
-      this.forceWrite();
-      // Verify ownership after the rename to guard against two processes
-      // racing on the same stale lock. Only the process whose PID is now
-      // in the file actually won the race.
+      if (!this.reclaimStaleLock()) return false;
       const verify = this.read();
       if (!verify || verify.pid !== process.pid) return false;
       this.acquired = true;
@@ -160,11 +157,35 @@ export class CuratorLock {
     }
   }
 
-  private forceWrite(): void {
-    const body = JSON.stringify({ pid: process.pid, start_ts: this.now() });
+  private reclaimStaleLock(): boolean {
     const tmp = this.target + '.' + process.pid + '.tmp';
-    fs.writeFileSync(tmp, body, { mode: 0o600 });
-    fs.renameSync(tmp, this.target);
+    try {
+      fs.writeFileSync(tmp, JSON.stringify({ pid: process.pid, start_ts: this.now() }), {
+        flag: 'wx',
+        mode: 0o600,
+      });
+      if (!this.shouldReclaim()) return false;
+      try {
+        fs.unlinkSync(this.target);
+      } catch (e) {
+        const code = (e as NodeJS.ErrnoException).code;
+        if (code !== 'ENOENT') return false;
+      }
+      try {
+        fs.linkSync(tmp, this.target);
+        return true;
+      } catch {
+        return false;
+      }
+    } catch {
+      return false;
+    } finally {
+      try {
+        fs.unlinkSync(tmp);
+      } catch {
+        /* ignore */
+      }
+    }
   }
 
   private shouldReclaim(): boolean {
--- a/src/skill-memory/curator-lock.ts
+++ b/src/skill-memory/curator-lock.ts
@@ -1,0 +1,181 @@
+/**
+ * Curator single-instance lock (#715 v2).
+ *
+ * The curator runs as a background task; multiple `oc serve`
+ * processes on the same machine must not run it concurrently. Per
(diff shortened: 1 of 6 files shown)


shaun0927:
@codex review

Addressed the prior round of feedback in the most recent commit. Please re-review:

- The summary of changes is in the commit message.
- Tests pass locally (`npx jest tests/skill-memory` green; counts vary by PR).

If anything still warrants a P1/P2, please flag it; otherwise a 👍 is fine to signal merge-ready.

shaun0927:
Re: P2 "Reject non-object args payloads" on `src/perception/voting/providers/http-helpers.ts:100`

That file isn't modified by this PR (#763 only touches `src/skill-memory/curator-lock.ts`, `src/skill-memory/curator.ts`, and the curator-lock test). The voting provider's `asActionInvocation` was introduced earlier in the stack (PR #759 / #760). Treating it as in-scope for the curator PR widens the diff in a way that complicates review of the curator changes themselves.

The finding is valid in spirit and will be addressed in a follow-up PR scoped to the voting providers. The other two findings (P1 stale-lock reclaim atomicity and P2 `failWindowMs` plumbing) are addressed in commit `a750b36`:

- `curator-lock.ts:88-93` — after `forceWrite()`, re-read the lock file and verify `pid === process.pid` before declaring acquisition; if a foreign PID is in the file we lost the rename race and return `false`. Regression test added at `tests/skill-memory/curator-lock.test.ts` ('stale-lock race: second acquire loses when a different PID wins the rename').
- `curator.ts` — `SkillStatsResolver` signature extended with optional `windowMs`, and `statsResolver(rec)` is now `statsResolver(rec, failWindowMs)` so `OPENCHROME_CURATOR_FAIL_WINDOW_MS` is no longer a dead knob.

shaun0927:
Codex P1 remediation evidence for current head ee2d5e044bc8af0a3de6d0f2ee83f018851a44f6:

- Fixed stale lock reclaim in commit ee2d5e044bc8af0a3de6d0f2ee83f018851a44f6: `acquire()` now reclaims with an unlink-then-link path where only one process can create the target, followed by ownership verification. The old `forceWrite()` rename-over-target path is gone.
- Fixed demotion sidecar reset in the same commit: `demoteSkill()` now rewrites the sidecar to `count: 1` and a single recent run matching the reset frontmatter, so stale pre-demotion successes cannot immediately re-promote the skill.

Validation run after the fix:
- `npx jest tests/skill-memory/curator-lock.test.ts tests/skill-memory/curator.test.ts --runInBand` -> 23 passed
- `git diff --check` -> passed
- `npm run build` -> passed

I re-queried current-head bot P1 comments after pushing; the two old inline comments still map to the new head, but the code paths above and regression tests show they are stale against ee2d5e0.

shaun0927:
Replaced by PR #814 (merged to develop).

PR #93, opened 2026-02-26
fix: add missing safeTitle() calls and global tool execution timeout

shaun0927:
## Summary

Closes #92

The #65 fix (commit `5233f8e`) introduced `safeTitle()` to prevent `page.title()` hangs on frozen renderers, but **missed 2 call sites** and left no global safety net. This caused:

- `tabs_context_mcp` hanging 5+ minutes (N tabs × 30s per frozen tab)
- All interaction tools (`click_element`, `form_input`, etc.) adding 30s latency via `dom-delta.ts`
- Any tool handler able to hang the MCP server indefinitely

### Changes

| Priority | File | Fix |
|----------|------|-----|
| **P0** | `src/tools/tabs-context.ts` | `page.title()` → `safeTitle(page)` |
| **P0** | `src/utils/dom-delta.ts` | `page.title()` → `safeTitle(page)` |
| **P1** | `src/mcp-server.ts` | Wrap `tool.handler()` in 120s `Promise.race` timeout |
| **P2** | `src/cdp/client.ts` | Add 5s timeout to `target.page()` fallback path |
| **P3** | `src/index.ts` | Add `unhandledRejection` / `uncaughtException` handlers |
| — | `src/config/defaults.ts` | Add `DEFAULT_TOOL_EXECUTION_TIMEOUT_MS` constant |

### Why the previous navigate fix wasn't enough

The #65 fix correctly addressed cookie scan timeouts and added `safeTitle()` to `navigate.ts`, `page-reload.ts`, `tabs-create.ts`, and `wait-for.ts`. However:

1. **`tabs-context.ts` was missed** — this tool iterates ALL tabs across ALL workers, making the N×30s multiplication worst case
2. **`dom-delta.ts` was missed** — this runs after every interaction tool, adding 30s when post-navigation renderers are busy
3. **No architectural safety net** — a single missed timeout anywhere could still hang the entire MCP server indefinitely

This PR fixes all three gaps.

## Test plan

- [ ] `tsc --noEmit` passes (verified locally — 0 errors)
- [ ] `tabs_context_mcp` returns within 3s even with frozen tabs (safeTitle 3s timeout per tab)
- [ ] Interaction tools (`click_element`, etc.) don't add 30s delay after navigation
- [ ] Any tool that exceeds 120s returns a timeout error instead of hanging
- [ ] `target.page()` fallback returns null after 5s instead of blocking indefinitely
- [ ] Unhandled promise rejections are logged to stderr instead of silently crashing

[...]

Issue #92, opened 2026-02-26
fix: missing safeTitle() calls and global tool timeout cause 5-10min hangs


shaun0927:
## Problem

After the issue #65 fix (commit `5233f8e`), `safeTitle()` was introduced to prevent `page.title()` hangs on frozen renderers. However, **2 call sites were missed**, and there is **no global safety net timeout** at the MCP server level.

This causes:
- `tabs_context_mcp` hanging for 5+ minutes (iterates all tabs, each bare `page.title()` can block 30s)
- All interaction tools (`click_element`, `form_input`, etc.) adding 30s latency via `dom-delta.ts`
- Any tool handler hanging indefinitely with no upper bound

### BUG 1 (P0): `tabs-context.ts:64` — bare `page.title()` without timeout

```typescript
title: await page.title(),  // NO timeout — blocks up to 30s per tab
```

Worst case: 5 workers × 5 tabs = 25 tabs. If 10 are frozen: **10 × 30s = 300s (5 min)**.

`safeTitle()` was applied to `navigate.ts`, `page-reload.ts`, `tabs-create.ts`, `wait-for.ts` but **missed here**.

### BUG 2 (P0): `dom-delta.ts:283` — bare `page.title()` without timeout

```typescript
title = await page.title();  // NO timeout — blocks 30s on frozen renderer
```

DOM delta runs after nearly every interaction tool. Post-navigation renderers are commonly busy.

### BUG 3 (P1): No global tool execution timeout in `mcp-server.ts:429`

```typescript
result = await tool.handler(sessionId, toolArgs);  // NO timeout wrapper
```

If any tool handler hangs, the entire MCP request hangs indefinitely. No safety net.

### BUG 4 (P2): `cdp/client.ts:811` — `target.page()` fallback has no timeout

```typescript
const page = await target.page();  // can block indefinitely on bad target
```

After reconnection, `targetIdIndex` cache is cleared, forcing all lookups through this fallback path.

### BUG 5 (P3): No `unhandledRejection` handler in `src/index.ts`

Zero `unhandledRejection` / `uncaughtException` handlers in the entire `src/`. Background tasks (heartbeat, pool maintenance) can silently crash the process.

## Acceptance Criteria

- [ ] AC1: `tabs_context_mcp` uses `safeTitle(page)` instead of bare `page.title()`
- [ ] AC2: `dom-delta.ts` uses `safeTitle(page)` instead of bare `page.title()`
[...]

PR #88, opened 2026-02-26
feat: signal degraded mode in MCP tool responses (#76)

shaun0927:
## Summary

Closes #76.

When OpenChrome falls back to a temporary profile (because the real Chrome profile is locked by another process), **nothing in MCP tool responses currently indicates this degraded state**. This causes LLMs to:
- Retry authentication flows that can never succeed without real cookies
- File incorrect bug reports against websites
- Waste tokens on impossible recovery loops

This PR adds **profile-state signaling** so LLMs can detect degraded mode and adjust their strategy accordingly.

### 1. ProfileState tracking in ChromeLauncher (`src/chrome/launcher.ts`)

- Added `ProfileType` union type: `'real' | 'persistent-synced' | 'temp-snapshot' | 'temp-fresh'`
- Added `ProfileState` interface with fields: `type`, `cookieCopiedAt?`, `extensionsAvailable`, `sourceProfile?`, `userDataDir?`
- Track profile state at **5 decision points** in `ensureChrome()`:
  - Existing Chrome attachment → `real`
  - Explicit `userDataDir` → `real`
  - Real profile successfully unlocked → `real`
  - Profile locked, cookies copied → `temp-snapshot` (with `cookieCopiedAt` timestamp)
  - No profile found → `temp-fresh`
- Exposed `getProfileState()` method returning a shallow copy (immutable from outside)

### 2. `_profile` metadata injection in MCPServer (`src/mcp-server.ts`)

- Follows the existing `_timing` injection pattern — adds `_profile` to every tool response
- **Success path**: injects `_profile` object with `type` and `extensions` fields, plus `cookieAge` when applicable
- **Error path**: same `_profile` metadata (no warning text)
- **First-call warning**: on the first tool call with a non-`real` profile, prepends a human-readable warning to `content[]` explaining:
  - Current profile type and what it means
  - Which capabilities are missing (extensions, saved passwords, localStorage, etc.)
  - Actionable recommendation (e.g., "Close other Chrome instances and restart")
- Warning shown only once per session (`profileWarningShown` flag)
- Added `formatCookieAge()` helper for human-readable age strings ("5s ago", "3m ago", "2h ago")
- Added `oc_profile_status` to the `lifecycle` tool category

### 3. New `oc_profile_status` tool (`src/tools/profile-status.ts`)

- On-demand profile capability inspection tool
- Returns structured JSON with:
  - `profileType`: current profile type string
[...]

PR #1122, opened 2026-05-13
fix: clear title and timeout guard timers

shaun0927:
## Summary

Follow-up cleanup from the audit of #88, #91, and #92-related PRs (#93/#94):

- route `safeTitle()` through the shared `withTimeout()` helper so its timeout handle is cleared when `page.title()` completes quickly
- replace the remaining raw `page.title()` metadata collection call sites in checkpoint/session snapshot tools with `safeTitle(page)`
- clear losing timers in localized CDP timeout races
- replace DOM delta timeout races with `withTimeout()`
- add focused `safeTitle()` tests for success cleanup and timeout fallback

## Validation

- `npm run build:src` passed in Codex autopilot before final commit
- `npm test -- --runInBand --runTestsByPath tests/utils/safe-title.test.ts` passed
- TypeScript diagnostics for changed files: 0 errors
- `git diff --check` passed

Note: a combined local run including `tests/utils/with-timeout.test.ts` hit an existing wall-clock sensitivity on this busy Fly worker (`Date.now() - start < 500`) while the repo already had unrelated long-running `tsc`/`jest` jobs. The new `safeTitle` test itself passes quickly.

### DIFF
--- a/src/cdp/client.ts
+++ b/src/cdp/client.ts
@@ -1574,15 +1574,18 @@ export class CDPClient {
           );
         }
         if (cookieScan.targetId) {
+          let cookieCopyTid: ReturnType<typeof setTimeout> | undefined;
           await Promise.race([
             this.copyCookiesViaCDP(cookieScan.targetId, page),
-            new Promise<void>((resolve) =>
-              setTimeout(() => {
+            new Promise<void>((resolve) => {
+              cookieCopyTid = setTimeout(() => {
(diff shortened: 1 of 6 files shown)


shaun0927:
Addressing Codex P2 feedback: the observer injection timeout should remain a soft timeout that proceeds to the action rather than taking the no-delta fallback. I am preserving that behavior while still clearing the guard timer, then pushing a follow-up commit.

shaun0927:
Closing this PR because GitHub's PR head stayed pinned to the first commit even after the branch ref advanced. Re-opening the same final diff from a fresh branch so CI/review run against the reviewed head including the Codex feedback fix.

PR #1124, opened 2026-05-13
fix: preserve DOM delta soft timeout behavior

shaun0927:
## Summary

Follow-up to #1122 after Codex review feedback:

- preserve DOM delta observer injection's historical soft-timeout behavior
- still clear the injection guard timer when `page.evaluate(INJECT_OBSERVER_SCRIPT)` wins quickly
- keep the action/delta collection path available after a slow observer injection instead of falling back immediately to `delta: ''`

## Context

Codex correctly noted on #1122 that changing observer injection to `withTimeout()` made the 5s guard reject and take the no-delta fallback path. This PR restores the previous soft-timeout semantics while retaining the timer cleanup goal from the #88/#91/#92 audit follow-up.

## Validation

- `git diff --check` passed
- TypeScript diagnostics for `src/utils/dom-delta.ts`: 0 errors

### DIFF
--- a/src/utils/dom-delta.ts
+++ b/src/utils/dom-delta.ts
@@ -292,9 +292,20 @@ export async function withDomDelta<T>(
     preUrl = '';
   }
 
-  // Inject the MutationObserver
+  // Inject the MutationObserver. Preserve the historical soft-timeout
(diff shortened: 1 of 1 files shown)


shaun0927:
@codex review

PR #1276, opened 2026-05-15
fix(watchdog): don't auto-launch Chrome on startup readiness probe

nodejun:
## Summary
- Startup readiness probe (`wireChromeReadiness().initializeStartupConnection()`) called `cdpClient.connect()` without overriding `autoLaunch`, so each MCP server boot with `--auto-launch` (or global config `autoLaunch: true`) spawned a fresh Chrome window — even when no tool had been invoked yet. Visible symptom: every VSCode reload pops an empty browser.
- Threads `autoLaunch?: boolean` through `CDPClient.connect()` to `connectInternal()` (mirrors the existing `autoLaunch: false` pattern used on force-reconnect paths in `client.ts:589, 1044`), and pins the startup probe to `{ autoLaunch: false }`.
- Spawning stays where it belongs: actual tool-call code paths, where caller intent is explicit. No behavior change for `session-manager` / `connection-pool` callers — they still rely on `this.autoLaunch` from the constructor.

## Files changed
- `src/cdp/client.ts` — `connect()` accepts and forwards `autoLaunch` to `connectInternal()`
- `src/watchdog/chrome-readiness.ts` — startup probe calls `connect({ autoLaunch: false })` with a comment explaining intent
- `tests/watchdog/chrome-readiness.test.ts` — updated mock signature; new assertion `expect(client.connect).toHaveBeenCalledWith({ autoLaunch: false })`

## Test plan
- [x] `tsc -p tsconfig.json --noEmit` clean
- [x] `jest tests/watchdog/chrome-readiness.test.ts` — 4/4 pass (including new autoLaunch:false assertion)
- [x] `jest tests/cdp/client-contracts.test.ts tests/cdp/session-init-budget.test.ts tests/cdp/anti-hang-defenses.test.ts` — 20/20 pass
- [x] `jest tests/session-manager-evict-target.test.ts tests/session-manager-contracts.test.ts tests/mcp-server.test.ts` — 40/40 pass
- [x] **Real-world smoke test**: ran patched binary with `--auto-launch --port 9223` on a free port → server boots through readiness wiring → **no Chrome process spawned on 9223**. Before the patch, this exact invocation would spawn Chrome.
- [x] Reloaded VSCode against the patched build — no empty Chrome window on extension-host restart.

### DIFF
--- a/src/cdp/client.ts
+++ b/src/cdp/client.ts
@@ -924,8 +924,9 @@ export class CDPClient {
    * retry loop stays within the caller's deadline (A-3). Coalesced callers
    * share whatever budget the first caller supplied.
    */
-  async connect(options?: { budget?: Budget }): Promise<void> {
+  async connect(options?: { budget?: Budget; autoLaunch?: boolean }): Promise<void> {
     const budget = options?.budget;
+    const autoLaunch = options?.autoLaunch;
     if (this.browser && this.browser.isConnected()) {
       // Skip active probe if recently verified by heartbeat (avoids per-call overhead)
       if (Date.now() - this.lastVerifiedAt < DEFAULT_CONNECT_VERIFY_STALENESS_MS) {
@@ -969,7 +970,7 @@ export class CDPClient {
     const generation = ++this.connectionGeneration;
     const pendingConnect = (async () => {
       try {
-        const connected = await this.connectInternal({ budget, generation });
+        const connected = await this.connectInternal({ budget, autoLaunch, generation });
         if (connected === false) {
           return;
         }
--- a/src/cdp/client.ts
+++ b/src/cdp/client.ts
@@ -119,6 +119,14 @@ export class CDPClient {
   private lastCookieScanResult: CookieScanResult | null = null;
   /** Coalesces concurrent connect() calls — only one connectInternal() runs at a time. */
   private pendingConnect: Promise<void> | null = null;
+  /**
+   * Effective `autoLaunch` of the currently in-flight `pendingConnect`. Used to
+   * reject coalescing when a new caller's effective `autoLaunch` differs — e.g.
+   * the startup readiness probe must not silently ride a tool-call's launch
+   * attempt and inadvertently spawn Chrome, nor must a tool call silently ride
+   * the probe's `false` and fail to launch.
+   */
+  private pendingConnectAutoLaunch: boolean | undefined = undefined;
   /** Invalidates stale async connection attempts that resolve after a reconnect. */
   private connectionGeneration = 0;
   /** Timestamp of last successful connection verification (heartbeat or active probe). */
@@ -927,6 +935,11 @@ export class CDPClient {
   async connect(options?: { budget?: Budget; autoLaunch?: boolean }): Promise<void> {
     const budget = options?.budget;
     const autoLaunch = options?.autoLaunch;
+    // Resolve the effective autoLaunch the same way connectInternal() does so
+    // coalescing comparisons (below) are apples-to-apples — callers who would
+    // both end up with the same behavior coalesce, callers who would diverge
+    // never silently inherit each other's choice.
+    const effectiveAutoLaunch = autoLaunch ?? this.autoLaunch;
     if (this.browser && this.browser.isConnected()) {
       // Skip active probe if recently verified by heartbeat (avoids per-call overhead)
       if (Date.now() - this.lastVerifiedAt < DEFAULT_CONNECT_VERIFY_STALENESS_MS) {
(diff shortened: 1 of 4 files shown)


shaun0927:
## Code Review (3-agent pipeline: code-reviewer / silent-failure-hunter / platform-reviewer)

**Verdict**: `P0 = 0`, `P1 = 1`, `P2 = 5` — `⚠️ FIX one race before merge`.

The fix correctly threads `autoLaunch` through `connect()` → `connectInternal()` using `??` (not `||`), so the explicit `false` from the startup probe is preserved past the instance-default fallback. The watchdog suppression of Chrome spawning at boot is sound. However the promise-coalescing path silently undermines the very invariant this PR establishes — see P1 below.

---

### P1 — Must Fix

#### [P1] Promise coalescing discards the caller's `autoLaunch` override
**File**: `src/cdp/client.ts:964-966`

`connect()` coalesces concurrent calls by returning the already-in-flight `this.pendingConnect`. That pending IIFE was closed over the **first** caller's `autoLaunch` value. Two races result:

1. A tool-call `connect()` (which defaults to `this.autoLaunch === true`) is in flight when `initializeStartupConnection()` fires `connect({ autoLaunch: false })`. The probe joins the in-flight launch — **Chrome is spawned at boot anyway**, defeating the PR.
2. The reverse: the probe's `connect({ autoLaunch: false })` lands first. A real tool-call `connect()` joins it and **fails to launch Chrome** even though the user asked for it.

The window is narrow (server boot), but it is exactly the scenario the PR is designed to harden.

**Suggested fix**: track the in-flight `autoLaunch` and only coalesce when the values match. Sketch:

```ts
private pendingConnectAutoLaunch?: boolean;
// ...
if (this.pendingConnect && this.pendingConnectAutoLaunch === autoLaunch) {
  console.error('[CDPClient] Coalescing concurrent connect() call');
  return this.pendingConnect;
}
this.pendingConnectAutoLaunch = autoLaunch;
// ... existing IIFE ...
```

Mismatched callers fall through and start a fresh attempt rather than silently inheriting the wrong `autoLaunch`.

---

### P2 — Improve (follow-up acceptable)

#### [P2] `forceReconnect` does not forward `autoLaunch` from the caller
**File**: `src/cdp/client.ts:953-956`

The probe-failure path falls into `await this.forceReconnect({ budget })` without forwarding `autoLaunch`. It happens to be safe today because `forceReconnect` internally hardcodes `autoLaunch: false` at line 1045, but the invariant is incidental. A future change to `forceReconnect` that lets it inherit `this.autoLaunch` would silently re-introduce the spawn-at-boot bug. Either thread `autoLaunch` into `forceReconnect`'s signature, or add a one-line comment at line 955 making the load-bearing behavior explicit.

#### [P2] Fragile microtask-count flush in the failure test
**File**: `tests/watchdog/chrome-readiness.test.ts:65-66`

`await Promise.resolve(); await Promise.resolve();` works today because the `.catch()` is exactly two microtasks deep. A future implementation hop will silently make this test pass without ever entering the catch body. Replace with the idiom used at line 48:

```ts
await client.connect.mock.results[0].value.catch(() => {});
```

#### [P2] Missing coverage: `connect()` resolves without emitting a `connected` event
**File**: `tests/watchdog/chrome-readiness.test.ts`

The "already-connected, recently-verified" shortcut in `CDPClient.connect()` returns immediately without firing a connection event. No test asserts that the readiness state then remains untouched (it is driven by the listener, not `connect()` resolution). Worth one additional test to lock the contract.

#### [P2] `ChromeReadinessClient` interface widening
**File**: `src/watchdog/chrome-readiness.ts:7`

The watchdog interface now exposes `{ autoLaunch?: boolean }`, which is only meaningful to the launching implementation. Minor layering smell — a JSDoc note clarifying that no-op implementations may safely ignore the option would suffice.

#### [P2] Self-documenting test assertion
**File**: `tests/watchdog/chrome-readiness.test.ts:44-45`

A one-line comment above the new assertion (mirroring the one added in `chrome-readiness.ts:67-72`) would tie the test to the invariant it protects.

---

### Things the review explicitly confirmed are fine

- The `??` (nullish-coalescing) guard at `connectInternal` line 676 — `false` is preserved, not collapsed to the instance default. Correct.
- No platform-portability issues: zero `process.env.HOME`, `/dev/tty`, hardcoded paths, or unguarded signals across the three files.
- Coalescing aside, no swallowed CDP errors / unhandled promise rejections introduced by the change.
- Existing non-watchdog callers (`session-manager.ts:348,350`, `connection-pool.ts:75`) are unaffected — they still pick up `this.autoLaunch` from the constructor.

---

Once the P1 coalescing fix lands, this PR is a clean, well-scoped bugfix and a UX win for users running with `--auto-launch` / `autoLaunch: true`. The P2s are non-blocking polish.

shaun0927 (approved):
Approving — the P1 coalescing race surfaced in the earlier review is fixed (now correctly tracks `pendingConnectAutoLaunch` and refuses to coalesce mismatched callers), the P2 follow-ups (forceReconnect intent comment, interface JSDoc, robust test flush, additional regression tests) are all in, and the cross-platform CI matrix (ubuntu/macos/windows × Node 18/20/22) is green.

PR #1406, opened 2026-05-27
feat(tools): oc_profile_fingerprint MCP tool (B3-PR2)

shaun0927:
## Summary

Expose `fingerprintEnvelope` (B3-PR1, #1387) as a **Tier-1 read-only MCP tool**. Captures the current tab's storage state (cookies + localStorage + sessionStorage + origin) via the same CDP walker `oc_context_export` uses, runs it through the deterministic fingerprint pipeline, and returns:

```ts
{ hash, version, algorithm, breakdown }
```

The hash is a **shape hash** — values never enter it. Two sessions with identical layouts but different secrets fingerprint identically; the negative tests on `fingerprintEnvelope` enforce that invariant.

**Stacked on top of #1387 (B3-PR1).** Base branch is `feat/b3-fingerprint`.

## Why

Per #1359 §Pillar B (profile/auth reuse) + §Pillar D (portable memory): the host uses this tool to ask *"is this the same logged-in session?"* without needing access to the secrets themselves.

## Wiring

- new `READ_ONLY` entry in `TOOL_ANNOTATIONS`;
- registered in `registerAllTools()` with capability `'core'`;
- v1.11 tools-list snapshot extended to include the new tool name.

## Defaults

`includeSessionStorage: false` so the fingerprint stays comparable across reattach (sessionStorage is ephemeral per tab). Callers that need to bind to a single live tab pass `true`.

## #1359 decision-rule check

| Rule | Answer |
|---|---|
| Pillar | B (profile/auth reuse), D (portable memory) |
| stdio/HTTP | both — read-only tool surface |
| Third-party credentials at boot | none |
| LLM judgment in host | yes — hash is a label, host decides |
| Real Chrome state | read-only (captures cookies/storage; never writes) |
| Portable artifact | yes — JSON `{hash, version, algorithm, breakdown}` |

## Test plan

- [x] `npx tsc --noEmit -p tsconfig.json` — clean
- [x] `npx jest tests/storage-state/fingerprint.test.ts tests/capability-filter.test.ts` — 14 + snapshot tests passing (no regression on the pure function or the tools-list baseline)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1359, opened 2026-05-25
SSOT: OpenChrome as a host-neutral MCP browser harness


shaun0927:
## Status — living SSOT (2026-07-29)

This issue defines durable product direction; it is not a sprint backlog. It is intentionally open, assigned, labeled `epic`, `status:tracking`, and `stale-exempt`, and should be updated when direction changes rather than closed for inactivity.

Normative resolved decisions live in `docs/roadmap/ssot-decisions.md`. Implementation work lives in focused issues with acceptance criteria and milestones.

| Area | State | Canonical tracking |
| --- | --- | --- |
| Host-neutral MCP browser harness identity | **Landed / ongoing invariant** | This issue + `docs/roadmap/portability-harness-contract.md` |
| Single-owner broker, auto-elect, leases, queues, diagnostics | **Landed** | `docs/roadmap/ssot-decisions.md` D3; v1.12.9 |
| Package update vs. MCP host-config migration | **Landed** | #1489 closeout; #1490 and #1491 |
| Lazy Chrome launch for ordinary stdio hosts | **Landed on `main`** | #1528 |
| Trustworthy bounded nightly E2E | **Landed on `main`** | #1571 |
| `oc_assert` canonical `contract_id` registry wiring | **Landed on `main`** | #1570 |
| Durable `evidence_handle` / `trace_ref` | **Landed on `main`** | #1572 / #1596 |
| Performance and console assertion/evidence kinds | **Landed on `main`** | #1569 / #1597 |
| Competitor pin, standalone binaries, explicit tab activation | **Landed on `main`** | #1554, #1555, #1556 |
| Daemon-backed playbook validation against session-visible tools | **Blocked on `main` session-scoped disclosure** | #1578; prerequisite #1573 remains `develop`-only |
| Server-side LLM planning, hidden host-specific behavior, unverified memory auto-execution | **Rejected** | Explicit non-goals below |

---

## Purpose

This issue is the product-direction SSOT for OpenChrome.

OpenChrome should remain a **host-neutral MCP browser harness for real Chrome**: useful from Codex, Claude Code, Hermes Agent, OpenClaw, IDEs, CI jobs, dashboards, and custom MCP clients through the same MCP contract.

[...]

PR #1407, opened 2026-05-27
feat(storage-state): portable signed snapshot envelope v1 (B3-PR3)

shaun0927:
## Summary

Introduce `src/storage-state/envelope.ts` — wraps an `EnvelopeCapture` into a transport-portable, host-verifiable record:

```ts
{ version, fingerprint, captured_at, origin, payload, hmac? }
```

**Stacked on top of #1406 (B3-PR2 fingerprint tool).** Base branch is `feat/b3-fingerprint-tool`.

## Two composable verification layers

1. **Fingerprint** (B3-PR1) — secret-free shape hash. Lets two parties compare envelopes without inspecting the payload.
2. **Optional HMAC-SHA256** over the canonical envelope encoding (excluding the `hmac` field itself). Detects tampering of the payload *values* that the shape hash deliberately doesn't cover.

The HMAC key is **host-supplied** — neither the envelope module nor the broader openchrome boot path ever generates, stores, or requires an HMAC key. Compliant with #1359 §P7 (no mandatory third-party credentials at boot).

## API

```ts
sealEnvelope(capture, opts?) → PortableSnapshotEnvelope
verifyEnvelope(raw, opts?) → { ok, envelope } | { ok: false, reason }
```

Closed `VerifyFailureReason` union: `unsupported_version | fingerprint_mismatch | hmac_missing | hmac_mismatch | hmac_key_missing | malformed`.

## Forward compatibility

A single `version` integer pins the canonical shape. Any change requires bumping the version; v1 envelopes remain readable forever.

## Why HMAC alongside fingerprint

The fingerprint is a *shape* hash — values never enter it. That's the intended secrecy property (B3-PR1 negative test). But it means same-length value tamper passes unsigned verification. HMAC catches that. The test `same-length value tamper passes unsigned (per value-secrecy invariant) but is caught when an HMAC is present` codifies this boundary.

## #1359 decision-rule check

| Rule | Answer |
|---|---|
| Pillar | B (profile/auth reuse), D (portable memory) |
| Third-party credentials at boot | none — HMAC key host-supplied, optional |
| stdio/HTTP | both — pure functions |
| LLM judgment in host | yes |
| Portable artifact | yes — versioned JSON |
| Constant-time HMAC compare | yes (`crypto.timingSafeEqual`) |

## Test plan

- [x] `npx tsc --noEmit -p tsconfig.json` — clean
- [x] `npx jest tests/storage-state/envelope.test.ts tests/storage-state/fingerprint.test.ts` — 31/31 passing
  - 18 new envelope tests across seal (5), verify happy path (2), verify failure modes (7), secrecy boundary (2), algorithm constants (2)
  - 14 inherited fingerprint tests still pass

### DIFF
--- a/src/storage-state/envelope.ts
+++ b/src/storage-state/envelope.ts
@@ -1,0 +1,221 @@
+/**
+ * Portable signed snapshot envelope — v1 (B3-PR3 of #1359).
+ *
+ * Wraps an `EnvelopeCapture` (cookies + localStorage + sessionStorage +
+ * origin) in a transport-portable record:
+ *
+ *   { version, fingerprint, captured_at, origin, payload, hmac? }
+ *
+ * Two parts compose the envelope:
+ *
+ *   1. The **fingerprint** (B3-PR1) — a secret-free hash of the
+ *      payload's shape. Lets two parties compare envelopes without
+ *      decrypting/inspecting the payload.
+ *
+ *   2. An optional **HMAC** over the canonical encoding of the
+ *      envelope (excluding the `hmac` field itself) so the recipient
+ *      can detect tampering. The key is **host-supplied** — neither
+ *      the envelope module nor the broader openchrome boot path ever
+ *      generates, stores, or requires an HMAC key. This keeps the
+ *      module compliant with #1359 §P7 (no mandatory third-party
+ *      credentials at boot).
+ *
+ * Pure functions, no I/O. The envelope can travel through MCP tool
+ * input/output, evidence bundles, or external benchmark adapters
+ * without code-mobility tricks — it's just JSON.
+ *
+ * Forward compatibility: a single `version` integer pins the canonical
+ * shape. Any change to the shape requires bumping the version; v1
+ * envelopes remain readable forever.
+ *
+ * @see docs/storage-state/fingerprint-spec.md
+ */
+
+import { createHmac, timingSafeEqual } from 'node:crypto';
+
+import {
+  fingerprintEnvelope,
+  type ProfileFingerprint,
+} from './fingerprint';
+import type { EnvelopeCapture } from './storage-state-manager';
+
+/** Envelope format version. */
+export const ENVELOPE_VERSION = 1 as const;
+
+/** HMAC algorithm used when the host supplies a signing key. */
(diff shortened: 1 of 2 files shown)


shaun0927:
# PR Review Summary

## Verdict
Approve

## Scope Reviewed
- PR intent: introduce portable snapshot envelope v1 with fingerprint verification and optional HMAC.
- Main changed areas: `src/storage-state/envelope.ts`, envelope tests.
- #1359 fit: aligned with Pillar B/D and §P7. HMAC key is host-supplied, no boot credential is required, and the artifact is portable JSON.
- Tests reviewed: seal/verify happy paths, tamper handling, HMAC failure modes, value-secrecy invariant, fingerprint tests, typecheck.
- Checks considered: GitHub reports no checks for this stacked branch; local targeted verification passed.

## Blocking Issues
Fixed in this PR.

### src/storage-state/envelope.ts — unsupported version reason was unreachable
Problem: `VerifyFailureReason` and PR text documented `unsupported_version`, but the structural guard classified version mismatch as `malformed` before reaching that branch.
Why it matters: hosts branch on precise failure reasons for trust policy; declared reasons should be reachable and stable.
Suggested fix: allow structurally valid unknown-version envelopes through the guard and return `unsupported_version` explicitly.
Suggested test: updated wrong-version test to assert `unsupported_version`.

## Warnings
None.

## Mutation-Test Thinking
- Likely mutants that should be killed: HMAC computed including its own field, value tamper with HMAC passing, missing HMAC accepted under requireHmac, wrong version collapsing to malformed.
- Mutants current tests may not catch: canonical payload ordering across future nested additions; acceptable for v1 fixed shape.

## Complexity / CRAP-style Risk
- High-risk functions/modules: security-adjacent verifier; tests cover important branches.
- Complexity increase: moderate but contained in pure functions.

## Security / Operational Risk
Acceptable after fix. Unsigned envelopes still cannot detect same-length value tamper by design; tests document that HMAC is the layer for value integrity.

## Verification
- `npm test -- --runInBand tests/storage-state/envelope.test.ts tests/storage-state/fingerprint.test.ts` passed 31/31.
- `npx tsc --noEmit -p tsconfig.json` passed.

## Final Recommendation
Merge-ready after fix commit `7db50d37`.

PR #1408, opened 2026-05-27
feat(tools): oc_context_export/import accept portable signed envelope (B3-PR4)

shaun0927:
## Summary

Wire the B3-PR3 portable envelope into both halves of the `oc_context_*` surface. Strictly additive: the existing `ContextEnvelope` output and `envelope` input are unchanged.

**Stacked on top of #1407 (B3-PR3 portable envelope).** Base branch is `feat/b3-portable-envelope`.

## Export — opt-in `signedEnvelope`

New args:
- `signedEnvelope: boolean` (default false)
- `hmacKey: string` (optional, only used when `signedEnvelope=true`)

When `signedEnvelope=true`, the response gains a `signed_envelope` field carrying the B3-PR3 `PortableSnapshotEnvelope` sealed over the same capture, fingerprinted, and optionally HMAC-signed.

## Import — `signed_envelope` accepted as an alternative

When supplied, the import:
1. **Verifies the portable envelope first** (fingerprint + optional HMAC via `hmacKey`). On failure returns `{ ok: false, integrityError: 'signed_envelope verification failed: <reason>' }` **without applying any state**.
2. Materializes a `ContextEnvelope` from the verified payload via `buildEnvelope()` so the existing strict-replace apply path stays the SSOT.

`signed_envelope` wins when both are supplied. `tabId` remains the only `required` field on import (envelope-or-signed-envelope is now the implicit constraint, surfaced as an error result rather than at the schema level so we can give a precise message).

## #1359 decision-rule check

| Rule | Answer |
|---|---|
| Pillar | B (profile/auth reuse), D (portable memory) |
| Third-party credentials at boot | none — HMAC key host-supplied, optional |
| stdio/HTTP | both — pure JSON I/O |
| Backward compatibility | strictly additive — existing `envelope`/`signedEnvelope=false` paths untouched |
| LLM judgment in host | yes — host picks signed vs unsigned |
| Portable artifact | yes — versioned JSON, fingerprint-verifiable, optionally HMAC-signed |

## Public surface

`src/storage-state/index.ts` re-exports fingerprint and envelope symbols so downstream consumers don't deep-import.

## Test plan

- [x] `npx tsc --noEmit -p tsconfig.json` — clean
- [x] `npx jest tests/storage-state` — 54/54 across 4 suites
  - 18 envelope tests (seal/verify/failure-modes)
  - 14 fingerprint tests
  - 22 inherited (storage-state-manager etc.)

### DIFF
--- a/src/storage-state/index.ts
+++ b/src/storage-state/index.ts
@@ -9,3 +9,23 @@ export {
   captureContextEnvelopeData,
   applyContextEnvelopeData,
 } from './storage-state-manager';
+
+export {
+  fingerprintEnvelope,
+  FINGERPRINT_VERSION,
+  FINGERPRINT_ALGORITHM,
+  type ProfileFingerprint,
+  type FingerprintBreakdown,
+} from './fingerprint';
+
+export {
+  sealEnvelope,
+  verifyEnvelope,
+  ENVELOPE_VERSION,
+  ENVELOPE_HMAC_ALGORITHM,
+  type PortableSnapshotEnvelope,
+  type VerifyResult,
+  type VerifyFailureReason,
+  type SealOptions,
+  type VerifyOptions,
+} from './envelope';
--- a/src/tools/oc-context.ts
+++ b/src/tools/oc-context.ts
@@ -456,6 +456,13 @@ const importDefinition: MCPToolDefinition = {
           '`signed_envelope`. Host-supplied; openchrome never stores ' +
           'or generates this key.',
       },
+      requireHmac: {
+        type: 'boolean',
+        description:
+          'When true, signed_envelope import requires a valid HMAC and ' +
+          'fails before applying state if the envelope is unsigned or no ' +
+          'hmacKey is provided. Default false.',
+      },
     },
     required: ['tabId'],
   },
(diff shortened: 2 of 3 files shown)


shaun0927:
# PR Review Summary

## Verdict
Approve

## Scope Reviewed
- PR intent: wire portable signed envelopes into `oc_context_export/import` additively.
- Main changed areas: storage-state exports, `oc_context` export/import args and verification path, oc_context regression tests.
- #1359 fit: aligned with Pillar B/D and §P7. Core supplies portable mechanism; host supplies HMAC key and policy.
- Tests reviewed: oc_context strict-replace/security tests, signed envelope failure path, envelope/fingerprint tests, typecheck.
- Checks considered: GitHub reports no checks for this stacked branch; local targeted verification passed.

## Blocking Issues
Fixed in this PR.

### src/tools/oc-context.ts — import lacked an MCP-level requireHmac switch
Problem: the portable verifier supports `requireHmac`, and the trust-policy docs require hosts to enforce it in multi-tenant flows, but `oc_context_import` did not expose the option.
Why it matters: MCP hosts need an import-time enforcement knob so unsigned signed-envelope payloads can fail before any cookie/storage mutation.
Suggested fix: added `requireHmac` input and threads it into `verifyEnvelope`; added a regression test proving unsigned signed_envelope with `requireHmac:true` fails before state changes.
Suggested test: `oc-context-replace.test.ts` signed-envelope no-apply assertion.

## Warnings
None.

## Mutation-Test Thinking
- Likely mutants that should be killed: applying state before verification, ignoring requireHmac, accepting wrong HMAC, legacy envelope path regressing.
- Mutants current tests may not catch: full live browser export/import of signed envelopes; unit harness covers the apply gate and shared storage walker.

## Complexity / CRAP-style Risk
- High-risk functions/modules: auth-state import path; mitigated by verification-first control flow and no-apply regression test.
- Complexity increase: moderate, localized.

## Security / Operational Risk
Acceptable after fix. HMAC key remains host-supplied and not persisted; import failure short-circuits before mutation.

## Verification
- `npm test -- --runInBand tests/tools/oc-context-replace.test.ts tests/storage-state/envelope.test.ts tests/storage-state/fingerprint.test.ts` passed 53/53.
- `npx tsc --noEmit -p tsconfig.json` passed.

## Final Recommendation
Merge-ready after fix commit `7113d198`.

Issue #1373, opened 2026-05-27
feat: make lifecycle and reconnect broker-aware and lease-aware


shaun0927:
## Parent

Part of #1359.

## Problem

In shared-profile multi-client mode, reconnect, shutdown, memory-pressure cleanup, and `oc_stop` must be coordinated by the single owner/broker. If every client treats disconnect or cleanup as a local browser lifecycle event, active sessions can lose tabs or tools unexpectedly.

## Goal

Make browser lifecycle and reconnect broker-aware and lease-aware.

## Scope

- Separate client disconnect from browser shutdown.
- Ensure client stdio disconnect does not close shared Chrome while active leases remain.
- Make reconnect a broker-level operation observed by clients rather than repeated independently by each client.
- Make `oc_stop` broker-aware and lease-aware.
- Ensure memory-pressure cleanup uses lease priority and cleanup policy.
- Persist enough state to recover target ownership after transient CDP reconnect.
- Add crash/reconnect tests for owner, client, Chrome, stale target, and lease expiry scenarios.

## Non-goals

- Designing the full broker transport from scratch.
- Replacing existing single-client lifecycle behavior before broker mode is available.
- Guaranteeing recovery from unrecoverable Chrome/profile corruption.

## Suggested PR split

1. Client disconnect versus browser lifecycle separation.
2. Broker-level reconnect state and client notification/resource reporting.
3. Lease-aware `oc_stop` and memory-pressure cleanup.
4. Crash/reconnect scenario tests.

## Acceptance criteria

- One client disconnecting does not close tabs or drop tools for other active clients.
- Broker reconnect preserves active leases when Chrome targets still exist.
- Cleanup cannot close another active session's leased target except under a documented emergency policy.
- Reconnect state is visible to clients in a host-neutral MCP-compatible way.

shaun0927:
Resolved by #1416 (Track broker lifecycle for reconnect-safe shared stops) and #1420 (Record browser lane target restoration state). Closing.

PR #1416, opened 2026-05-27
Track broker lifecycle for reconnect-safe shared stops

shaun0927:
## Summary\n- add process-local broker lifecycle state for owner/reconnect/stop diagnostics\n- expose broker lifecycle in oc_connection_health\n- make oc_stop default to keepChrome in broker owner mode unless explicitly overridden\n\n## Issues\n- Advances #1373\n- Aligns with #1359 by protecting the single shared Chrome owner from accidental proxy/client shutdown semantics\n\n## Validation\n- npm test -- tests/broker-lifecycle.test.ts --runInBand\n- npm run build\n- git diff --check\n\n## Notes\n- Live crash/reconnect validation still requires an operator Chrome environment; this PR locks the state model and safety default.

### DIFF
--- a/src/broker/lifecycle.ts
+++ b/src/broker/lifecycle.ts
@@ -1,0 +1,83 @@
+export type BrokerLifecycleMode = 'disabled' | 'owner';
+export type BrokerLifecycleState = 'idle' | 'running' | 'reconnecting' | 'stopping' | 'stopped';
+
+export interface BrokerLifecycleSnapshot {
+  mode: BrokerLifecycleMode;
+  state: BrokerLifecycleState;
+  startedAt: string | null;
+  updatedAt: string | null;
+  reconnectCount: number;
+  lastEvent: string | null;
+  ocStopDefaultKeepChrome: boolean;
+}
+
+const lifecycle: BrokerLifecycleSnapshot = {
+  mode: 'disabled',
+  state: 'idle',
+  startedAt: null,
+  updatedAt: null,
+  reconnectCount: 0,
+  lastEvent: null,
+  ocStopDefaultKeepChrome: false,
+};
+
+function nowIso(): string {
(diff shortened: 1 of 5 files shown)


shaun0927:
## Review summary — merge-ready after wire-in fix

### Intent vs #1359
- Advances #1373 by tracking process-local broker owner lifecycle (`mode`, `state`, `reconnectCount`, `lastEvent`) and surfacing it via `oc_connection_health.brokerLifecycle`.
- Aligns with #1359 **P3a** by protecting the shared Chrome owner from accidental proxy/client shutdown semantics.

### Gap found and fixed
The PR body promised `oc_stop` defaults to `keepChrome=true` in broker owner mode, but `shouldOcStopKeepChromeByDefault()` was exported and unit-tested without being wired into `src/tools/shutdown.ts` — the handler kept the old hard-coded `|| false` fallback. Pushed a 5-line fix (commit `2779d12a`) that:
- Imports the helper.
- Uses it only when the caller did **not** pass `keepChrome` explicitly (explicit `true`/`false` still wins, so operators can force a full stop).
- Updates the schema description so MCP clients see the broker-mode default.

Tests: `npm test -- tests/broker-lifecycle.test.ts tests/mcp-server-shutdown.test.ts --runInBand` → 7/7 pass; `npm run build` clean.

### Remaining scope
`recordBrokerReconnect('reconnecting'|'reconnected'|'reconnect_failed')` is intentionally a state surface; the CDP reconnect callsite wiring is documented as a follow-up in the PR's "Not-tested" line (live broker crash/reconnect against a real Chrome process). The lifecycle state is observable + safe-by-default already, which matches the PR body's contract.

### CI
All completed jobs pass; macOS pending = runner queue. Latest commit will re-trigger CI.

### Verdict
Merge-ready after the wire-in fix lands green.

PR #1418, opened 2026-05-27
Reuse startup NTP for first default-context page

### LINKED ISSUE #1346: Fix extra blank startup NTP tab left after tabs_create
## Symptoms

When `tabs_create` is called (and also `navigate` / `crawl` / `crawl_sitemap` / `batch_paginate` / `validate_page`, which go through the first `createTarget` path), a **blank tab (`chrome://newtab/` or `about:blank`) remains open alongside the requested URL tab**. The prune logic sometimes closes it, but depending on the race it can remain visible or briefly flicker.

### 1) Where does the startup NTP tab come from?
- `spawn(chromePath, args, ...)` in `src/chrome/launcher.ts:458-496` does **not pass any URL argument**.
- When Chrome starts with only `--no-restore-last-session` and `--no-first-run`, it opens one NTP (`chrome://newtab/` or `chrome://new-tab-page/`) by default.
- In other words, immediately after startup Chrome already contains one “untracked NTP” tab.

### 2) `createPage` ignores the NTP and creates another new tab
- The default-context branch (`context == null`) in `src/cdp/client.ts:1640-1648` always calls `browser.newPage()`.
- Result: **1 blank tab + 1 requested tab = 2 tabs**.
- `useConnectionPool: true` is the default, but the pool’s `createNewPage()` eventually calls `cdpClient.createPage(undefined, undefined, true)` as well (`src/cdp/connection-pool.ts:363`), so it follows the same problematic path.

### 3) The prune logic tries to clean this up afterward, but it is fragile
- `src/session-manager.ts:987-1101` contains startup blank-tab cleanup logic with a 500 ms `setTimeout`.
- The guard `existingTargetIds.size === 1` is too strict: it can be false if pool prewarming is active or if Chrome has not exposed the NTP target yet immediately after startup.
- `about:blank ↔ chrome://newtab/` URL transition race: if the snapshot saw `about:blank` and the cleanup still sees the same value, `!cleanupExistingIds.has(...)` becomes false, so the tab cannot be closed.
- The 500 ms delay is best-effort by nature, so the response can return first and the user may see flicker.

### In scope

**File**: `createPage` in `src/cdp/client.ts` (default-context branch only)

At the start of the `context == null` branch, reuse the startup NTP when all four guards below pass:

| Guard | Meaning |
|---|---|
| `context == null` | Automatically excludes named contexts (#848) |
| `getChromeLifecycleMode() === 'isolated'` | Protects user tabs in attach mode |
| `targetIdIndex.size === 0` | Only the true first call per `CDPClient` (safe for multi-Chrome scenarios) |
| Exactly one untracked page candidate from `browser.targets()`, with URL in `{NTP, about:blank, ''}` | Automatically falls back for restore-last-session or multi-tab startup cases |

When the gate passes, obtain the page via `target.page()` and reuse it as-is. After that, keep the existing flow unchanged: `configurePageDefenses`, `applyRegisteredPreloads`, `setViewport`, and `smartGoto` should still run as before.

### Out of scope (separate PR)
- Reusing NTP in `createTargetStealth` (separate code path)
- Simplifying the prune logic (it still owns Site Isolation orphan blank-tab cleanup)
- Passing an initial URL to the launcher
- Redesigning pool prewarming (default `min=0`, so this is unaffected)

## Test Plan
- First `createPage` (managed isolated + default context): reuse the NTP and keep `browser.targets()` page count at 1
- Second `createPage`: create a new tab and increase page count to 2
- Attach mode: reuse must not trigger
- Named context: reuse must not trigger; the page must be created inside the named context
- Abnormal startup cases with 0 or 2 pages: fall back to `browser.newPage()`
- `validate_page` regression: ensure `url=undefined` calls do not depend on the startup URL

## Expected Impact
- No remaining blank tabs after `tabs_create` / `navigate`
- No visible flicker, because the root cause is removed instead of relying on cleanup afterward
- Deterministic behavior independent of the prune logic

shaun0927:
## Summary\n- reuse the single startup blank/NTP target on the first default-context createPage call\n- keep guardrails for attached Chrome, non-first calls, isolated contexts, and multiple candidates\n- add CDP client contract tests for reuse and non-reuse paths\n\n## Issues\n- Fixes #1346\n- Aligns with #1359 browser-visible stability/polish without changing shared-profile ownership semantics\n\n## Validation\n- npm test -- tests/cdp/client-contracts.test.ts --runInBand\n- npm run build\n- git diff --check

### DIFF
--- a/src/cdp/client.ts
+++ b/src/cdp/client.ts
@@ -1709,7 +1709,11 @@ export class CDPClient {
         await smartGoto(page, url, { timeout: DEFAULT_NAVIGATION_TIMEOUT_MS });
         assertDomainAllowed(page.url());
       } catch (err) {
-        // Close the page to prevent about:blank ghost tabs on navigation failure
+        // Close the page to prevent about:blank ghost tabs on navigation failure.
+        // When `page` came from findReusableStartupPage the closed tab is the
+        // startup NTP — that's intentional: reuse is gated to isolated-mode
+        // (OpenChrome owns Chrome), so the user does not see a stuck about:blank
+        // from a failed first navigation.
         const targetId = getTargetId(page.target());
         this.targetIdIndex.delete(targetId);
         await page.close().catch(() => {});
--- a/src/cdp/client.ts
+++ b/src/cdp/client.ts
@@ -1638,14 +1638,21 @@ export class CDPClient {
         }),
       ]) as Page;
     } else {
-      // Create page in Chrome's default context
-      let newPageTid2: ReturnType<typeof setTimeout>;
-      page = await Promise.race([
-        browser.newPage().finally(() => clearTimeout(newPageTid2)),
-        new Promise<never>((_, reject) => {
-          newPageTid2 = setTimeout(() => reject(new Error(`newPage() timed out after ${DEFAULT_NEW_PAGE_TIMEOUT_MS}ms`)), DEFAULT_NEW_PAGE_TIMEOUT_MS);
-        }),
-      ]) as Page;
+      const startupPage = await this.findReusableStartupPage(browser);
+      if (startupPage) {
+        page = startupPage;
+        skipCookieBridge = true;
+        console.error(`[CDPClient] Reused startup tab ${getTargetId(page.target())} for first default-context page`);
+      } else {
+        // Create page in Chrome's default context
+        let newPageTid2: ReturnType<typeof setTimeout>;
+        page = await Promise.race([
+          browser.newPage().finally(() => clearTimeout(newPageTid2)),
+          new Promise<never>((_, reject) => {
+            newPageTid2 = setTimeout(() => reject(new Error(`newPage() timed out after ${DEFAULT_NEW_PAGE_TIMEOUT_MS}ms`)), DEFAULT_NEW_PAGE_TIMEOUT_MS);
+          }),
+        ]) as Page;
+      }
 
       // Copy cookies from an authenticated page (skip for pool pre-warming to avoid
       // CDP session conflicts and unnecessary overhead on about:blank pages).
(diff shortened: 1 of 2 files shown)


shaun0927:
### Review verdict: merge

**Intent and direction.** First-call reuse of the startup blank/NTP tab — directly maps to #1359 Pillar P3 (real Chrome polish: "no surprise blank tabs") and #1346.

**Alignment with #1359 SSOT.**
- `findReusableStartupPage` gates reuse on `getChromeLifecycleMode() === 'isolated'`, so attach mode (user-owned Chrome) is never touched. ✅ Pillar B attach-mode safety.
- `targetIdIndex.size !== 0` + `candidates.length !== 1` give a robust "exactly first call, exactly one matching target" guard. ✅ Pillar B deterministic tab lifecycle.
- The reused page is treated like a fresh one (configurePageDefenses, applyRegisteredPreloads, setViewport, smartGoto). `configurePageDefenses` is idempotent (`__defensesConfigured` guard), `applyRegisteredPreloads` uses a `WeakMap<Page, Set<string>>` so it is also idempotent. ✅
- `skipCookieBridge = true` for the reused NTP is correct — chrome://newtab has no auth state to inherit, and bridging from there would find nothing useful.

**Polish in [f9df7c01](https://github.com/shaun0927/openchrome/commit/f9df7c01):**
- Added a third client-contract test asserting that after the first default-context page is recorded in `targetIdIndex`, a second `createPage` call falls through to `browser.newPage()` instead of re-consuming an NTP-shaped target. Previously only the launchMode guard had test coverage.
- Annotated the navigation-failure close block to document that closing a reused startup NTP on failed first navigation is intentional (isolated-mode only, OpenChrome owns Chrome → no user-visible tab is destroyed).

**CI.** ubuntu × {18,20,22} and windows × {18,20,22} green; macos jobs queued.

Ready to merge.

Issue #1037, opened 2026-05-12
feat(harness): task-scoped browser lanes for isolated parallel work


shaun0927:
## Why

Goose subagents isolate work so the main agent context is not polluted by every exploratory step. OpenChrome should not spawn LLM subagents, but it can safely adopt the isolation principle at the browser harness layer: **bounded browser lanes** for host-driven parallel work.

OpenChrome already has sessions, workers, target IDs, tenant management, and tab pools. This issue makes that isolation explicit and task-aware so a host agent can run parallel browser branches without cross-contaminating refs, budgets, traces, or recovery state.

Related work:
- Builds on the task envelope issue and #855 task ledger.
- Complements existing `SessionManager`, tenant registry, and CDP connection pool.
- Does not replace team/swarm/host-agent orchestration.

## Scope

Add task-scoped browser lanes. A lane is a named isolated browser execution branch owned by a task, backed by existing session/worker/tab primitives.

### Tool surface

```ts
oc_lane_create({ taskId, name?, purpose?, initialUrl?, budget? })
oc_lane_list({ taskId })
oc_lane_get({ taskId, laneId })
oc_lane_close({ taskId, laneId })
```

Existing browser tools may accept optional `laneId` alongside `taskId`. When supplied, refs and tab lookup should resolve within that lane first.

### Lane state

```ts
interface BrowserLane {
  lane_id: string;
  task_id: string;
  name?: string;
  purpose?: string;
  status: 'open' | 'closing' | 'closed' | 'failed';
  sessionId: string;
  workerId: string;
  targetIds: string[];
  created_at: number;
  last_activity_at: number;
  counters: { toolCalls: number; failures: number };
}
```

## Non-goals

- Do not spawn LLM subagents.
- Do not add autonomous planning, voting, or task decomposition.
- Do not create a new Chrome process per lane unless the existing session/tenant policy already requires it.
- Do not make `laneId` mandatory for existing tools.

## Implementation notes

- Prefer extending `SessionManager`/worker mapping rather than duplicating target ownership.
- Ref IDs generated inside a lane must not accidentally resolve to another lane when `laneId` is provided.
- Lane close should clean up lane-owned tabs and release ledger state, but must not close unrelated task/session tabs.
- Lane state should be visible in `oc_task_get` and trace metadata.
- Apply existing idle eviction semantics; lanes should not keep dead tabs alive forever.

## Acceptance criteria

[...]

PR #1420, opened 2026-05-27
Record browser lane target restoration state

shaun0927:
## Summary\n- add per-target lane status metadata for open vs target_missing\n- add reconciliation helper to mark restored lanes with missing targets as recoverable failures\n- cover restart-style missing-target behavior in lane tests\n\n## Issues\n- Advances #1037\n- Aligns with #1359 by keeping browser lanes host-directed and deterministic\n\n## Validation\n- npm test -- tests/core/browser-lanes.test.ts --runInBand\n- npm run build\n- git diff --check

### DIFF
--- a/src/core/browser-lanes.ts
+++ b/src/core/browser-lanes.ts
@@ -46,10 +46,14 @@ function laneWithStatuses(lane: BrowserLane, liveTargetIds?: Set<string>): Brows
     status: liveTargetIds && !liveTargetIds.has(targetId) ? 'target_missing' as const : 'open' as const,
   }));
   const missing = statuses.some((target) => target.status === 'target_missing');
+  // A reconciled lane with zero targets is treated as recoverable failure too:
+  // hosts need to see "lane has no live targets" rather than a silent open status.
+  const emptyAfterReconcile = liveTargetIds !== undefined && statuses.length === 0;
+  const degraded = missing || emptyAfterReconcile;
   return {
     ...lane,
-    ...(statuses.length ? { targetStatuses: statuses } : {}),
-    ...(missing ? { status: 'failed' as const, recovery: 'target_missing' as const } : {}),
+    targetStatuses: statuses,
+    ...(degraded ? { status: 'failed' as const, recovery: 'target_missing' as const } : {}),
   };
 }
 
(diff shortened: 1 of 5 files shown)


shaun0927:
## Merge-readiness review (post-fix)

**Verdict**: merge-ready after the wiring fix in `7c2ca2af`.

### #1359 alignment
- **P2 (harness, not agent)**: lane reconciliation only reports facts; hosts decide whether to recreate. ✓
- **P3a (single CDP owner / lease ownership)**: lane state stays a thin task-ledger overlay on existing SessionManager primitives. ✓
- **P4 (facts before decisions)**: per-target `target_missing` status is exposed rather than auto-recovered. ✓
- No new outbound LLM calls, no mandatory client capability, fully backward-compatible (`targetStatuses?`, `recovery?` are optional). ✓

### Fix landed in this iteration
The first review found a P0: `reconcileBrowserLaneTargets` was implemented and unit-tested but **never called from any production path**, so the acceptance criterion ("hosts can see which lanes lost their targets after restart") was unmet at runtime.

Commit `7c2ca2af` wires the helper into `oc_session_resume`:
- `collectLiveTargetIds(tabAnalysis)` aggregates live target IDs from the existing tab analysis (LIVE/REMAPPED `currentTargetId`) plus a fallback enumeration of `SessionManager.getAllSessionInfos() × getWorkerTargetIds()`.
- `reconcileLanesForResume(taskId, tabAnalysis)` calls `reconcileBrowserLaneTargets` when a `taskId` is supplied; failures are swallowed so the rest of the resume guide stays useful.
- `generateResumeGuide` now renders a `Browser lanes:` section showing `lane_id`/`name`/`status`/`recovery=target_missing`/`missing=N/M`, plus a one-line recovery hint when any lane is degraded.
- 3 new tests cover the live-id collection helper (LIVE+REMAPPED, SessionManager fallback, disconnected SessionManager) and 2 new `generateResumeGuide` tests cover the lane rendering branches.

### Validation
- `npm run build` ✓
- `npx jest tests/tools/session-resume.test.ts tests/core/browser-lanes.test.ts --runInBand` → 52 passed, 0 failed
- CI matrix on PR pre-fix push was already green; awaiting re-run on the wiring commit.

### Suggested follow-up (non-blocking)
Consider a lower-level lifecycle hook (e.g. on CDP reconnect) so reconcile runs even when no host explicitly calls `oc_session_resume`. Acceptable to defer per #1037 since the explicit-resume path is the documented host workflow.

Issue #876, opened 2026-05-12
feat(tools): sampling-driven disambiguation PoC for act (mcp-servers adoption 7)


shaun0927:
## Motivation (narrow PoC)

`src/tools/act.ts` (natural-language action API, #578) needs to disambiguate when multiple DOM elements match the user's intent (e.g., "click the submit button" when the page has 3 submit-like elements). Today the heuristic is server-side: heuristic ranking by aria-label, role, visibility. This is fragile and grows with every reported failure case.

MCP spec defines `sampling/createMessage` — server-to-client LLM round-trip. The client's LLM evaluates the ambiguity (DOM snippet + intent + screenshot) and returns a selection. The server **does not** make any outbound LLM API call — the round-trip is fully routed through the client's transport to the client's chosen LLM provider.

### Portability-Harness Contract compliance
- **P1 (tool server identity):** Sampling is a per-tool-call assistance round-trip, not orchestration. The server is still answering a single `act` invocation.
- **P3 (anywhere-compatible MCP):** No outbound LLM API call from the server; no API keys required; pure spec mechanism. Compliant.

This is a **narrowly scoped PoC** to validate the pattern before considering it for `vision-find`, captcha kind detection, or `smart-goto` fallback selection. Those are explicitly **deferred** until this issue closes with measured wins.

### Hard scope limits (non-negotiable for this PR)
- Only `act` tool uses sampling.
- Only the "multiple candidates after deterministic ranking" branch invokes sampling.
- Fallback to current heuristic if `capabilities.sampling` not advertised.
- No new dependencies; no outbound network beyond CDP/test URLs.

### Implementation
- `src/tools/act.ts`:
  - When 2+ candidates remain after current heuristic, AND the connected client advertised `sampling`:
    1. Build a `sampling/createMessage` request with:
       - System prompt: "You are choosing one DOM element to satisfy the user's intent. Reply with JSON: `{\"index\": N, \"confidence\": 0..1, \"reason\": \"...\"}`. No prose."
       - User message: candidate list (selector + text + role + bbox), original intent string, attached image (screenshot of region).
       - `modelPreferences: { intelligencePriority: 0.7, speedPriority: 0.3 }`.
       - `maxTokens: 200`.
    2. Send via `src/mcp-server.ts` `requestSampling(...)` helper (new — accepts originating sessionId, routes via active transport).
    3. Parse response strict-JSON; on parse failure, fall through to heuristic.
[...]

PR #1421, opened 2026-05-27
Add client-mediated sampling fallback to act

shaun0927:
## Summary\n- add MCP client-mediated sampling refinement for multi-step act sequences when the client advertises sampling\n- strictly parse sampled JSON actions and fall back to deterministic parsed actions on timeout/error/malformed output\n- surface sampling metadata without adding server-side LLM calls\n\n## Issues\n- Advances #876\n- Aligns with #1359 by keeping sampling client-mediated and deterministic by default\n\n## Validation\n- npm test -- tests/tools/act-sampling.test.ts --runInBand\n- npm run build\n- git diff --check

### DIFF
--- a/src/tools/act.ts
+++ b/src/tools/act.ts
@@ -163,12 +163,20 @@ async function maybeRefineActionsWithSampling(
         },
       }],
       maxTokens: 400,
-    }, { timeoutMs: 5000, signal: context.signal });
+    }, { timeoutMs: 8000, signal: context.signal });
     const sampled = parseSampledActions(response);
     if (!sampled) return { actions, decision: { used: false, supported: true, fallbackReason: 'invalid_sampling_response' } };
     return { actions: sampled, decision: { used: true, supported: true } };
   } catch (err) {
-    return { actions, decision: { used: false, supported: true, fallbackReason: err instanceof Error ? err.message : String(err) } };
+    // Map known transport/cancel signatures to closed-set reasons so we don't
+    // leak raw transport text to clients in `_meta.sampling.fallbackReason`.
+    const message = err instanceof Error ? err.message : String(err);
+    const fallbackReason = /timeout/i.test(message)
+      ? 'timeout'
+      : /abort|cancel/i.test(message)
+        ? 'cancelled'
+        : 'transport_error';
+    return { actions, decision: { used: false, supported: true, fallbackReason } };
   }
 }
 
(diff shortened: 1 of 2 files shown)


shaun0927:
## Merge-readiness review (post-fix)

**Verdict**: merge-ready after the timeout/sanitization fix in `09cfd032`.

### #1359 alignment
- **P1 (host-neutral)**: deterministic fallback intact when client lacks `sampling` capability or `requestClient`. ✓
- **P2 (harness, not agent)**: zero outbound server-side LLM calls; sampling is a client-mediated MCP round-trip. ✓
- **P4 (facts before decisions)**: parsed deterministic actions are the safe default; sampling only refines after deterministic ranking. ✓
- **P7 (core boring)**: sampling gated on `source === 'parsed'` — cached/templated sequences are never re-sampled. ✓

### Stage-1 spec check
- `VALID_ACTIONS` whitelist matches the 9-verb `ActionVerb` union exactly.
- `requestClient(method, params, options)` matches `ToolContext.requestClient`'s `<T>(string, Record<string, unknown>?, { timeoutMs?, signal? }?) => Promise<T>` signature.
- `_meta.sampling` envelope only added when the client opted in.

### Fixes landed in this iteration
Two MEDIUM findings from initial review:

1. **Timeout drift from #876 spec** — bumped `timeoutMs` from `5000` to `8000` so it matches the issue's 8s client-mediated round-trip budget. Sampling latency that crosses two IPC boundaries can comfortably exceed 5s on first call.
2. **`fallbackReason` leaked raw transport text** — added a closed-set mapping `{timeout | cancelled | transport_error | invalid_sampling_response | sampling_unavailable}` so `_meta.sampling.fallbackReason` is portable across hosts and stable for clients to act on, instead of surfacing internal transport strings.

Added two new tests: one for `cancelled` (AbortError-like), one for `transport_error` (opaque connection failure). The existing timeout test now asserts the closed-set `timeout` value.

### Deferred to follow-up (non-blocking)
- Issue #876 also lists an `audit_logger` `sampling_request` event. This PR ships the fallback path and `_meta` surface; the audit-log subsystem hook is a separate concern and can land in a follow-up.
- Integration test exercising `coreHandler` with `source='cache'` to confirm sampling skip would strengthen coverage; the gate itself is a one-line guard with the inline comment justifying it.

### Validation
- `npm run build` ✓
- `npx jest tests/tools/act-sampling.test.ts --runInBand` → 8/8 passed
- Pre-fix CI matrix was already 9/9 green; awaiting re-run on `09cfd032`.

Issue #1431, opened 2026-05-28
feat(skill): fresh-lane re-verification gate before promoting a skill to recall


shaun0927:
## Why

Webwright addresses the "premature done" failure mode (model claims success but didn't actually finish) by **re-running the final script in a fresh folder** as a self-reflection completion gate.

OpenChrome already has the building blocks — \`oc_assert\` outcome contracts (#784), \`oc_skill_replay\` (#856), and the broker / lease / per-target queue stack landed via the parallelism PRs (#1370–#1373). What is missing is a **promotion step** between \`oc_skill_record\` and \`oc_skill_recall\`: a skill should not become recall-discoverable until it has been re-executed against its contract in a clean lane.

## Scope

- Add a \`promotion\` state to the skill memory store: \`recorded → re-verified → recallable\`.
- New internal procedure: claim a fresh browser lane (clean profile, no shared state from the recording lane), replay the skill via \`oc_skill_replay\`, gate on the original \`oc_assert\` contract.
- On verification failure, the skill stays in \`recorded\` and is excluded from \`oc_skill_recall\` results.
- Diagnostic surface so a host can observe which skills were quarantined and why.
- Plays into the SSOT (#1359) direction: directly exercises the lease registry (#1371) and per-target queue (#1372) on a real workflow.

## Acceptance

- [ ] Skill store carries a promotion state.
- [ ] \`oc_skill_recall\` only returns promoted skills unless an opt-in flag is set.
- [ ] Promotion runs against a freshly leased lane.
- [ ] Failure metrics surfaced via journal / evidence bundle.
- [ ] Integration test where a flaky skill is correctly quarantined.

shaun0927:
Implemented and merged to develop: promotion state (`recorded → re_verified → recallable`/`quarantined`) in skill-memory store, `oc_skill_recall` promotion filter with `include_quarantined` opt-in + legacy auto-fallback, fresh-lane promotion gate (`src/core/contracts/promotion-gate.ts`, `src/core/skill-memory/promote.ts`), and tests (`promotion-state.test.ts`, `promote.test.ts`, `promotion-gate.test.ts`). PRs #1439 #1447 #1448. All acceptance criteria met.

PR #1447, opened 2026-05-28
feat(skill): add promotion state to the skill memory store

shaun0927:
## Summary
Stacked on #1439 (Part 1).
- New `promotionState` field on `SkillRecord`: `recorded → re_verified → recallable`, plus terminal `quarantined`.
- `SkillMemoryStore.setPromotionState(skillId, state, at, reason?)` rotates the field. Caller supplies `at` for deterministic tests.
- `oc_skill_recall` filters by promotion state with safe defaults; new flags `include_unpromoted` and `include_quarantined` for opt-in.
- Legacy records normalise to `recorded` on read — recall semantics for v1.x data unchanged unless promotion is actively used.

## SSOT (#1359) alignment
- Pure store + recall change. No host coupling.

## Test plan
- [x] `tests/core/skill-memory/promotion-state.test.ts` — 7/7 pass (defaults, rotation, quarantine truncation, unknown-id throw, three recall filter modes).

Part 2 of #1431. Depends on #1439. Part 3 (fresh-lane re-verification procedure) follows.

### DIFF
--- a/src/core/skill-memory/store.ts
+++ b/src/core/skill-memory/store.ts
@@ -423,6 +423,17 @@ export class SkillMemoryStore {
       if (existing?.lastReplayError !== undefined) {
         next.lastReplayError = existing.lastReplayError;
       }
+      // Preserve promotion-lifecycle fields across idempotent re-record
+      // (#1431 Part 2). The recorder owns `steps`/`contractId`; the
+      // promotion procedure (Part 3) owns the state machine. Mixing them
+      // would silently demote a `recallable` skill back to invisible.
+      if (existing?.promotionState !== undefined) {
+        next.promotionState = existing.promotionState;
+        next.promotionStateAt = existing.promotionStateAt;
+        if (existing.promotionQuarantineReason !== undefined) {
+          next.promotionQuarantineReason = existing.promotionQuarantineReason;
+        }
+      }
       file.skills[skillId] = next;
       await this.writeFile(file);
       return { skill_id: skillId, stored_at: Date.now() };
@@ -602,12 +613,17 @@ export class SkillMemoryStore {
         ...existing,
         promotionState: nextState,
         promotionStateAt: at,
-        ...(nextState === 'quarantined' && quarantineReason
-          ? { promotionQuarantineReason: quarantineReason.slice(0, 512) }
-          : nextState === 'quarantined'
-            ? {}
-            : { promotionQuarantineReason: undefined }),
       };
+      if (nextState === 'quarantined') {
+        if (quarantineReason) {
+          next.promotionQuarantineReason = quarantineReason.slice(0, 512);
+        }
+      } else {
+        // Explicit delete keeps the in-memory object consistent with the
+        // on-disk JSON; relying on `JSON.stringify` to drop an `undefined`
+        // value would survive a future serializer change.
+        delete next.promotionQuarantineReason;
+      }
       file.skills[skillId] = next;
       await this.writeFile(file);
     } finally {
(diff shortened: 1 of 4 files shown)


shaun0927:

**SSOT (#1359) alignment — confirmed.** Pure skill-memory store + recall change, no host coupling, no model calls. Advances Pillar D (portable verified memory) and P6 (verified memory only): promotion gates recall on contract-verified re-runs while keeping the recall path LLM-free.

**Independent review surfaced 1 P1 + 3 P2; all fixed in `e666434f`:**

- **[P1] Legacy skills silently invisible after upgrade.** The default filter hid `recorded`, but legacy records normalize to `recorded` on read — so any v1.x store would return zero recall results until a promotion sweep ran, contradicting the "v1.x semantics unchanged" claim. **Fix:** `oc_skill_recall` now auto-falls-back to v1.x semantics when a domain has no `re_verified`/`recallable` skill; the recorded-vs-promoted filter only activates once at least one skill is promoted. Added `promotion_filter_active` to the output so callers can distinguish "no skills" from "skills exist but unpromoted".
- **[P2] Implicit erasure** of `promotionQuarantineReason` via `{ ...undefined }` spread → replaced with explicit `delete`.
- **[P2] `record()` dropped promotion fields** on idempotent re-record (silently demoting a `recallable` skill) → now preserved alongside the replay-outcome fields.
- **[P2] Test isolation:** `process.env.HOME` mutation did not redirect the in-process recall tool's store (and is non-portable: Windows reads `USERPROFILE`). Switched to mocking `os.homedir()`, which fixed a real-home pollution bug the original tests masked.

**Verification:** `npm run build` clean, `npm run lint` clean, promotion-state suite 11/11 pass (added quarantine→restore, re-record preservation, auto-fallback, and `promotion_filter_active` cases). No real-home pollution.

> Note: this PR targets the stacked base `feat/1431-scratch-profile-lane` (#1439), so the repo CI workflow (which runs only on PRs into `main`/`develop`) does not trigger here; it will run once the base merges and this PR auto-retargets to `develop`.

shaun0927:
## Independent review pass (merge-readiness) — verified ✅

Re-reviewed against the #1359 SSOT and the #1431 intent, with a fresh independent reviewer (the earlier Codex pass hit usage limits and never ran).

**Intent & direction — sound.** Part 2 of #1431 adds the `recorded → re_verified → recallable` (+terminal `quarantined`) state to the skill store and gates `oc_skill_recall` on it. This is squarely on Pillar D (portable verified memory) and P6 (verified memory only): recall stays LLM-free and only contract-verified skills become discoverable by default.

**#1359 alignment — confirmed.** Pure store + recall change, no host coupling, no model calls, fail-safe defaults. The auto-fallback (recall keeps v1.x semantics until a domain has a promoted skill) preserves the host-neutral, no-surprise contract for existing deployments.

**Independent findings:** 0 P0, 1 soft-P1, 2 P2 — none data-loss / merge-blocking. Core filter logic (auto-fallback, quarantine hide, promoted-only default), idempotent re-record field preservation, and `setPromotionState` lock discipline all verified correct.

**Fix applied (`1194bb87`):** the one actionable item — `include_quarantined` is independent of the recorded-vs-promoted filter, so on an unpromoted domain (auto-fallback) passing it alone surfaces recorded + quarantined skills. Documented in the schema and locked with a dedicated test. The remaining P2s are stylistic and deferred.

**Verification:** `npm run build` clean · `tests/core/skill-memory/promotion-state.test.ts` 12/12 pass (added the auto-fallback + `include_quarantined` case) · `eslint src` clean · depcruise: no tier violations.

> Stacked on #1439 (Part 1). The repo CI workflow only runs on PRs into `main`/`develop`, so it triggers once the base merges and this PR retargets to `develop`.

Issue #856, opened 2026-05-12
feat(pilot): oc_skill_replay — deterministic CDP-step replay with oc_assert contract gate (mcp-browser-use adoption B)


shaun0927:
## Why

`oc_skill_record` (`src/tools/oc-skill-record.ts:1-211`) and `oc_skill_recall` (`src/tools/oc-skill-recall.ts:1-123`) shipped via #785 store and retrieve `{domain, name, steps[], contract_id, frozen_snapshot?}` records, but **there is no executor**. A host LLM that calls `oc_skill_recall` receives an opaque step list and must hand-translate every step back into individual MCP tool calls — the round-trip cost defeats the purpose of skill memory and there is no objective signal that a recall is still valid against the current page.

`mcp-browser-use` couples skill replay with a deterministic gate: re-execute the recorded CDP step sequence, then run a bound assertion contract. If the assertion fails, the skill is demoted in subsequent ranking. This converts the skill memory from "trust the LLM to interpret old steps" to "trust the contract: replay either passes or it doesn't, and a passing replay is the success criterion."

Outcome Contracts (#784, #792) and skill memory (#785) are already merged. The pilot tier graph executor (#820) treats the skill graph as a runtime — but does not perform recorded-step replay. The missing piece is a single tool that closes the loop: recall → replay → assert → demote-on-fail.

## What

Add a single new MCP tool `oc_skill_replay(skill_id, contract_id?)` in the pilot tier and a contract-gated demotion path in skill recall ranking.

**Boundary**:
- `src/pilot/skill/replay.ts` (new) — reads a recorded skill via the existing `SkillMemoryStore`, walks `steps[]`, executes each via the existing CDP client.
- `src/tools/oc-skill-replay.ts` (new) registered in `src/tools/index.ts` only when the pilot flag is on (matches the gate pattern from #820).
- `src/core/skill-memory/store.ts` extended with a `recordReplayResult(skill_id, { passed_at?, failed_at?, error? })` write path; existing recall ranker reads `last_replay_failed_at` to demote.
- Trace integration: every replay writes one `skill_replay` event to the active session's JSONL via the existing trace recorder.
- Pilot tier: gated by `--pilot` and `OPENCHROME_SKILL_REPLAY=1`. When closed, the module is not imported (P2 zero-impact extension).

## Contract

```ts
// src/pilot/skill/replay.ts
export interface SkillReplayInput {
  skill_id: string;          // 16-hex from oc_skill_record
[...]

PR #1448, opened 2026-05-28
feat(skill): add fresh-lane skill promotion procedure

### LINKED ISSUE #1431: feat(skill): fresh-lane re-verification gate before promoting a skill to recall

shaun0927:
## Summary
Stacked on #1447 (Part 2). Closes #1431 once stack lands.
- `promoteSkill(store, skillId, deps)`: rotates `recorded → recallable` on PASS, quarantines on STEP_FAIL/CONTRACT_FAIL/PRECONDITION_FAIL, infra-fault tolerant.
- Dependency-injected: tests cover every branch without launching Chrome. Production wiring binds `openScratchLane` to `createBrowserLane({ profile: 'scratch' })` (#1439) and `replayInLane` to `oc_skill_replay` (#856).

## SSOT (#1359) alignment
- No model calls. Pure orchestration of deterministic primitives.

## Test plan
- [x] `tests/core/skill-memory/promote.test.ts` — 8/8 pass (PASS path, two FAIL outcomes, lane-open infra fault, replay-throw infra fault, idempotent recallable, quarantined refusal, unknown id).

Part 3 of #1431. Depends on #1447 → #1439.

### DIFF
--- a/src/core/skill-memory/promote.ts
+++ b/src/core/skill-memory/promote.ts
@@ -1,0 +1,122 @@
+/**
+ * Fresh-lane skill promotion procedure (#1431 Part 3).
+ *
+ * Bridges Parts 1 (scratch-profile browser lanes) and 2 (promotion
+ * state on SkillMemoryStore) into a single host-callable routine.
+ *
+ * The procedure is deliberately dependency-injected: the production
+ * wiring will pass in a scratch-lane factory that produces a
+ * BrowserLane via `createBrowserLane({ profile: 'scratch' })` and a
+ * replay runner that drives `oc_skill_replay` against that lane. Tests
+ * pass deterministic fakes so the algorithm can be exercised without
+ * starting Chrome.
+ *
+ * Outcomes:
+ *   - replay PASS  → setPromotionState(skillId, 're_verified', now)
+ *                    and then setPromotionState(skillId, 'recallable', now)
+ *                    in the same call. The two-step rotation gives
+ *                    observers a transient `re_verified` checkpoint they
+ *                    can monitor without changing semantics.
+ *   - replay FAIL  → setPromotionState(skillId, 'quarantined', now, reason).
+ *
+ * SSOT (#1359) alignment: the procedure never calls a model. It only
+ * orchestrates the deterministic replay + contract check that already
+ * lives in oc_skill_replay (#856), in a clean profile guaranteed by
+ * Part 1 of this issue.
+ */
+import type { SkillMemoryStore } from './store';
+
+export interface PromoteSkillDeps {
+  /** Allocate a scratch lane and return an opaque handle. */
+  openScratchLane(): Promise<{ laneId: string; taskId: string }>;
+  /** Close the scratch lane (also rm-rfs its profile dir per #1431 Part 1). */
+  closeScratchLane(handle: { laneId: string; taskId: string }): Promise<void>;
+  /**
+   * Replay the recorded skill in the scratch lane and gate on its
+   * contract. PASS means both the steps and the contract held.
+   */
+  replayInLane(
+    handle: { laneId: string; taskId: string },
+    skillId: string,
+  ): Promise<{ outcome: 'PASS' | 'STEP_FAIL' | 'CONTRACT_FAIL' | 'PRECONDITION_FAIL'; reason?: string }>;
+  /** Wall-clock injection for deterministic tests. */
+  now(): number;
+  /** Optional logger; defaults to no-op. */
+  log?: (event: { stage: string; skillId: string; details?: unknown }) => void;
+}
+
+export type PromotionOutcome =
+  | { promoted: true; from: 'recorded' | 're_verified'; to: 'recallable' }
+  | { promoted: false; quarantined: true; reason: string }
+  | { promoted: false; quarantined: false; reason: string };
+
+export async function promoteSkill(
+  store: SkillMemoryStore,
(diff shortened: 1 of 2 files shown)


shaun0927:

**SSOT (#1359) alignment — confirmed.** Pure orchestration of deterministic primitives (scratch lane + `oc_skill_replay`); no model calls. Advances P2 (harness not agent) and P6 (verified memory only).

**Independent review surfaced 2 P1 + 4 P2; all fixed in `4c18a4a8`:**

- **[P1] Uncapped quarantine reason in the returned outcome.** The store caps the persisted reason at 512 chars, but the `PromotionOutcome.reason` returned to callers was the full string — a multi-KB replay error or DOM excerpt (possibly with PII) could leak through MCP responses. **Fix:** the returned reason is now capped to the same 512-char limit.
- **[P1] Two-write crash window.** The PASS path wrote `re_verified` then `recallable`; a crash between the two left the skill stranded at `re_verified` permanently. **Fix:** the durable record is written straight to `recallable` in a single store write, and the `re_verified` checkpoint is emitted as a **log event** (preserving observability without the crash hazard).
- **[P2] No-op `from`** invented `re_verified` when already `recallable` → now returns the real prior state.
- **[P2] `'quarantined' in r` tautology** (both failure variants carry the field) → tests narrow with `r.quarantined` after `!r.promoted`; added a doc note on the union.
- **[P2] `closeScratchLane` errors swallowed** silently (hides scratch-profile/disk leaks) → now surfaced via a `close_scratch_lane_failed` log event without changing the outcome.
- **[P2] Test portability:** inject `rootDir` instead of mutating `process.env.HOME` (Windows reads `USERPROFILE`).

**Verification:** `npm run build` clean, `npm run lint` clean, promote suite 10/10 pass (added reason-cap and close-failure cases). No real-home pollution.

> Stacked on #1447 → #1439; CI runs after the base chain merges and this PR auto-retargets to `develop`.

shaun0927:
## Independent review pass (merge-readiness) — verified ✅

Fresh independent review against #1359 SSOT and #1431 intent (the earlier Codex pass hit usage limits).

**Intent & direction — sound.** Part 3 of #1431 wires Parts 1+2 into a single host-callable `promoteSkill`: open a scratch lane, replay the skill against its contract, then PASS → `recallable` / fail → `quarantined`. This is the re-verification gate the issue asked for.

**#1359 alignment — confirmed.** No model calls; pure orchestration of deterministic primitives (scratch lane #1439 + `oc_skill_replay` #856). Advances P2 (harness not agent) and P6 (verified memory only). The dependency-injected shape keeps it portable and testable without Chrome.

**State machine verified correct:** infra faults (lane-open throw, replay throw, unknown id) correctly avoid quarantine and leave state untouched; PASS path is a single write to `recallable` (the `re_verified` checkpoint is log-only, eliminating the stranded-state crash window); scratch lane is always closed; quarantine reason capped before it can leak through MCP.

**Independent findings:** 0 P0, 2 P1, 2 P2 — none a data-loss risk for the (current) single-caller usage.

**Fixes applied (`12433a55`):**
- **Type clarity:** the two `PromotionOutcome` failure variants were structurally identical, so the union collapsed anyway and conveyed false precision → merged into one `{ promoted: false; quarantined: boolean; reason: string }` member, `quarantined` documented as the discriminant.
- **Concurrency contract:** `promoteSkill` snapshots state then awaits lane-open + replay before its final write, and `setPromotionState` has no compare-and-swap — concurrent promotion of the *same* skillId could clobber a terminal state. Documented that the host must serialize promotion per skill (the SSOT #1359 single-owner / per-target-queue model). Different skillIds remain safe to run concurrently.

The remaining P2s (log co-locating the outcome on close-fail; `'re_verified'` in the `from` union for legacy records) are stylistic/correct-as-is and deferred.

**Verification:** `npm run build` clean · `tests/core/skill-memory/promote.test.ts` 10/10 pass · `eslint src` clean.

> Stacked on #1447 → #1439; CI runs once the base chain merges and this PR retargets to `develop`.

PR #1438, opened 2026-05-28
feat(benchmark): Online-Mind2Web dataset loader and schema fixtures

shaun0927:
## Summary

- Adds `tests/benchmark/datasets/online-mind2web/loader.ts` exporting `OnlineMind2WebTask` type and `loadOnlineMind2Web()` function with `fixture` (CI-safe) and `hf` (opt-in network) modes
- Adds `tests/benchmark/datasets/online-mind2web/fixtures/sample-10.json` — 10-task fixture hand-crafted to match the published HF schema
- Adds unit tests (`loader.test.ts`): 14 passing, 1 correctly skipped (HF live fetch requires `OPENCHROME_OM2W_FETCH=1`)
- Adds `tests/benchmark/datasets/online-mind2web/README.md` with CC-BY 4.0 attribution paragraph
- Adds Online-Mind2Web row to `benchmark/COMPETITORS.md` registry (pinned dataset commit `7ab0fc3b5e0420f6a74c4e0f0faebc1f3eddb0c1`, measured-at: TBD-by-runner-PR)

## License attribution

Dataset: [osunlp/Online-Mind2Web](https://huggingface.co/datasets/osunlp/Online-Mind2Web), licensed CC-BY 4.0. Attribution is in both `README.md` and the top-level JSDoc block of `loader.ts`. The fixture file contains hand-crafted tasks that conform to the published schema; no raw dataset rows are reproduced.

## Files added

- `tests/benchmark/datasets/online-mind2web/loader.ts`
- `tests/benchmark/datasets/online-mind2web/loader.test.ts`
- `tests/benchmark/datasets/online-mind2web/fixtures/sample-10.json`
- `tests/benchmark/datasets/online-mind2web/README.md`
- `benchmark/COMPETITORS.md` (one row added)

## Test plan

- [x] `npx jest --testPathPattern="datasets/online-mind2web/loader.test.ts"` → 14 passed, 1 skipped
- [x] Fixture happy path: 10 tasks, each conforming to schema
- [x] Schema rejection: missing `task_id`, `website`, non-integer `reference_length`, non-array input all throw with field-named error messages
- [x] HF fetch guard: throws with `OPENCHROME_OM2W_FETCH` in message when env var unset
- [x] Cache hit: pre-seeded cache served without network
- [ ] HF live fetch (requires `OPENCHROME_OM2W_FETCH=1` + network) — skipped in CI, covered by the skip-guarded test

[...]

Issue #1427, opened 2026-05-28
benchmark: add Online-Mind2Web adapter to the live-LLM benchmark suite


shaun0927:
## Why

Webwright (Microsoft Research, 2026-05) reports headline numbers on **Online-Mind2Web** (300 live tasks across 136 sites, 100-step budget, LLM-as-a-Judge). The dataset is **publicly released under CC-BY 4.0** ([HF](https://huggingface.co/datasets/osunlp/Online-Mind2Web), [repo](https://github.com/OSU-NLP-Group/Online-Mind2Web), [paper](https://arxiv.org/abs/2504.01382)) and has an active leaderboard.

We already enforce equal-LLM / equal-budget headlines (#1310, #1424) and ship WebVoyager / real-world adapters under `benchmark/`. Adding Online-Mind2Web closes the gap with the current SOTA narrative and lets us publish directly comparable numbers.

## Scope

- New adapter under `benchmark/` consuming the HF dataset schema `{task_id, website, task_description, reference_length}`.
- Reproduce the reference evaluation: 100-step budget, automated LLM-as-a-Judge with screenshot+action-log evidence (mirroring the original eval settings as documented in the Webwright blog post).
- License attribution (CC-BY 4.0) wired into `benchmark/COMPETITORS.md` and generated reports.
- Integrate with the existing headline-eligibility gate (#1310) so partial / non-live runs cannot publish headline numbers.
- Equal-LLM cells for at least GPT and Claude families so we can claim parity numbers without depending on a single vendor.

## Acceptance

- [ ] Adapter loads the 300-task corpus from HF and runs the OpenChrome agent end-to-end.
- [ ] Judge module produces per-task verdicts + aggregate accuracy mirroring the published methodology.
- [ ] Generates a section in \`benchmark/BENCHMARK-REPORT.md\` gated by \`headline-gate.mjs\`.
- [ ] License + dataset card cited.

## Non-goals

- Reproducing Webwright's Odysseys numbers — that dataset is not publicly released. Tracked separately.

PR #1449, opened 2026-05-28
feat(benchmark): Online-Mind2Web runner shell with DI judge

shaun0927:
## Summary
Stacked on #1438 (Part 1).
- `runOnlineMind2WebTask(task, deps, opts)`: drives a single OM2W task through a step budget (default 100 — matches the published OM2W reference), captures evidence per step, forwards it to a pluggable LLM-as-Judge.
- Dependency-injected so the runner shape is reviewable without committing to a specific model. Production wiring will bind step() to the OpenChrome adapter and judge() to a pinned LLM prompt.
- Deterministic fake judge (`fakeSentinelJudge`) for CI smoke runs.

## SSOT (#1359) alignment
- No model calls in the runner itself — judge is host-provided.
- Pure additive change to the benchmark tree.

## Test plan
- [x] `tests/benchmark/datasets/online-mind2web/runner.test.ts` — 6/6 pass (default budget, sentinel pass, hard-fail stop, explicit budget, external shouldStop, judge_id forwarding).

Part 2 of #1427. Depends on #1438. Part 3 (headline-gate integration) follows.

### DIFF
--- a/tests/benchmark/datasets/online-mind2web/runner.test.ts
+++ b/tests/benchmark/datasets/online-mind2web/runner.test.ts
@@ -1,0 +1,90 @@
+/**
+ * Tests for the OM2W runner shell (#1427 Part 2).
+ *
+ * Drives the runner with deterministic fakes — no real LLM, no real
+ * browser. Validates step budget, early-stop, judge wiring.
+ */
+import {
+  runOnlineMind2WebTask,
+  fakeSentinelJudge,
+  type RunnerStep,
+  type RunnerDeps,
+} from './runner';
+import type { OnlineMind2WebTask } from './loader';
+
+function fakeTask(): OnlineMind2WebTask {
+  return {
+    task_id: 'om2w-fake-1',
+    website: 'https://example.com',
+    task_description: 'find the login link',
+    reference_length: 3,
+  };
+}
+
+function fakeStep(i: number, summary: string, ok = true): RunnerStep {
+  return { step: i, tool: 'navigate', args: { url: '/x' }, ok, summary };
+}
+
+describe('runOnlineMind2WebTask', () => {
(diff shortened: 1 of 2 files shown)


shaun0927:

**SSOT (#1359) alignment — confirmed.** No model calls inside the runner; the LLM-as-Judge is dependency-injected and host-provided. Advances P2 (harness not agent) and P5/Pillar E (benchmark-backed positioning). Default step budget 100 matches the published OM2W reference.

**Independent review surfaced 1 P1 + 3 P2; all fixed in `017500a1`:**

- **[P1] Unhandled step throw.** `await deps.step()` had no try/catch — a CDP crash, adapter error, or network timeout rejected the whole `runOnlineMind2WebTask` promise, so a batch harness lost the entire run instead of one task (inconsistent with the WebVoyager runner's structured-error contract). **Fix:** a thrown step is now caught, the judge rules on the partial evidence, and a structured failed `RunnerResult` (`reason: "step N threw: …"`) is returned.
- **[P2] `step_budget` edge cases.** `Infinity` passed the `> 0` guard and `Math.floor(Infinity)` → infinite loop; `0.5` floored to a silent zero-step no-op. **Fix:** floor first, then require `Number.isFinite && > 0`, else fall back to the default 100.
- **[P2] `judge_id` forwarding** used a truthy check that dropped an intentional empty-string id → now `!== undefined`.
- **[P2] `shouldStop` doc** referenced the wrong issue and implied it was inert → reworded to describe the live pre-step semantics.

**Verification:** `npm run build` clean, runner suite 10/10 pass (added throw-path, Infinity/fractional-budget, and immediate-shouldStop cases).

> Stacked on #1438 (loader); CI runs after the base merges and this PR auto-retargets to `develop`.

PR #1445, opened 2026-05-28
feat(contracts): add image_qa assertion to the outcome contract DSL

shaun0927:
## Summary
Stacked on #1441 (Part 1).
- New `image_qa` leaf assertion in the contract DSL: `{ question, expected_pattern }`.
- Evaluator goes through a new optional `EvalContext.imageQaSample` hook so the runtime can route to the `image_qa` MCP tool (and hence to host sampling). Absent hook or `unsupported_by_host` reply → inconclusive (passed:false).
- 10 tests: validator (4), evaluator (5), dispatch via `evaluate()` (1).

## SSOT (#1359) alignment
- No server-side LLM. Evaluator is inconclusive without host wiring.
- ReDoS guard via `validateRegexPattern` at author time **and** `compileSafeRegex` at evaluation time.

## Review-pass changes
- Documented the `image_qa` kind in `docs/contracts.md` and added it to the `oc_assert` `contract` schema description (hosts can now discover it).
- Routed the evaluator's runtime regex through `compileSafeRegex` for consistency/defense-in-depth.
- Added a validator test for a missing `expected_pattern`.

Part 2 of #1432. Depends on #1441.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

Issue #1432, opened 2026-05-28
feat(tools): image_qa primitive — vision Q&A over captured screenshots


shaun0927:
## Why

Webwright exposes a small but impactful tool, \`image_qa\`, that asks the configured model a focused question about a screenshot — used both for in-flight reasoning ("is the date picker actually open?") and for the self-reflection completion gate.

OpenChrome currently does not have a primitive in this shape. \`read_page\`, \`query_dom\`, and \`oc_observe\` cover structural inspection; nothing wraps a vision call against an already-captured screenshot. Adding this gives the host a cheap way to verify visual state without round-tripping through the agent's own context.

## Scope

- New MCP tool \`image_qa\` accepting \`{ screenshot_path | screenshot_ref, question, model? }\` and returning a structured verdict.
- Reuses the host's configured vision-capable model by default; opt-in override.
- Strict no-op without an explicit screenshot reference — never auto-captures, to keep the bill predictable.
- Host-neutral (SSOT #1359 compatible): the broker / lease layer is untouched; this tool only reads bytes that another tool already produced.
- Pairs with \`oc_assert\`: a contract can name an \`image_qa\` clause as one of its evidence sources.

## Acceptance

- [ ] New tool registered in \`src/tools/image-qa.ts\` with annotations and tier-1 wiring.
- [ ] Unit + integration tests against a fixture screenshot.
- [ ] Cost discipline: no auto-capture; the caller is responsible for providing the screenshot reference.
- [ ] Documentation under \`docs/\`.

shaun0927:
Implemented and merged to develop: `src/tools/image-qa.ts` + contract evaluator (`src/contracts/evaluators/image-qa.ts`) with annotations/tier-1 wiring, unit+integration tests (`tests/tools/image-qa.test.ts`, `tests/contracts/image-qa.test.ts`, `tests/contracts/oc-assert-image-qa-runtime.test.ts`), no-auto-capture cost discipline, and oc_assert pairing. PRs #1441 #1445 #1452 #1453. All acceptance criteria met.

PR #1452, opened 2026-05-28
feat(contracts): inject imageQaSample hook into oc_assert runtime

shaun0927:
## Summary
Stacked on #1445. Completes the runtime wire-up promised in the original #1432-B brief that was deferred from that PR.

- `image_qa.ts` exports a reusable `runImageQaSampling(ctx, params)` that does the sampling forwarding.
- `oc_assert.buildEvalContext` now accepts the calling `ToolContext` and injects an `imageQaSample` closure delegating to `runImageQaSampling`.
- `oc_assert` handler signature accepts the optional context parameter.
- `image_qa` evaluator stamps `details.error` on infra-fault paths so `isInconclusive` translates them to `verdict='inconclusive'` (not `fail`). Runtime wiring problems are not contract failures.

## Why this is a fix not a feature
Without this, the image_qa DSL clause was dead code in production: every contract returned `inconclusive` with reason `host_runtime_did_not_wire_imageQaSample`. PR #1445 implicitly required this wire-up to keep its commitment.

## SSOT (#1359) alignment
- No server-side LLM. The closure only forwards via `sampling/createMessage`, falls back to `unsupported_by_host`.

## Test plan
- [x] `tests/contracts/oc-assert-image-qa-runtime.test.ts` — 4/4 pass (pass on match, fail on mismatch, inconclusive without sampling cap, inconclusive without screenshot).
- [x] `tests/contracts/image-qa.test.ts` — 9/9 still pass.
- [x] `tests/tools/image-qa.test.ts` — 6/6 still pass.

Stacked on #1445 → #1441.

### DIFF
--- a/src/contracts/evaluators/image-qa.ts
+++ b/src/contracts/evaluators/image-qa.ts
@@ -10,6 +10,7 @@
  */
 import type { EvalContext } from '../eval-context';
 import type { EvaluationResult, ImageQaAssertion } from '../types';
+import { compileSafeRegex } from '../safe-regex';
 
 export async function evaluateImageQa(
   assertion: ImageQaAssertion,
@@ -22,6 +23,10 @@ export async function evaluateImageQa(
         passed: false,
         assertion_kind: 'image_qa',
         details: {
+          // `error` flips oc_assert verdict translation to inconclusive
+          // (not fail) — runtime wiring problems are not contract
+          // failures, they prevent evaluation entirely.
+          error: 'host_runtime_did_not_wire_imageQaSample',
           reason: 'host_runtime_did_not_wire_imageQaSample',
           question: assertion.question,
         },
@@ -37,6 +42,7 @@ export async function evaluateImageQa(
         passed: false,
         assertion_kind: 'image_qa',
         details: {
+          error: 'no_screenshot_available',
           reason: 'no_screenshot_available',
           question: assertion.question,
         },
@@ -56,6 +62,7 @@ export async function evaluateImageQa(
         passed: false,
         assertion_kind: 'image_qa',
         details: {
+          error: 'unsupported_by_host',
           reason: 'unsupported_by_host',
           host_reason: reply.reason,
           question: assertion.question,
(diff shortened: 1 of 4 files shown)


shaun0927:
## Analysis & merge summary (automated review)

**Intent / direction.** The runtime wire-up that *activates* the `image_qa` contract assertion from #1445. Correctly framed as a fix: without it, `EvalContext.imageQaSample` was never populated, so every `image_qa` contract evaluated to `inconclusive` in production. This injects the closure into `oc_assert.buildEvalContext` (forwarding to the host via `runImageQaSampling` → MCP `sampling/createMessage`) and stamps `details.error` on infra-fault paths so verdicts translate to `inconclusive` rather than `fail`.

**#1359 (SSOT) alignment — consistent.** The only outbound path is host-mediated MCP sampling — no `fetch`, no API key, no model SDK, no server-side judgment. When the client doesn't advertise `sampling` (or `requestClient` is absent), it returns `unsupported_by_host` and degrades deterministically to `inconclusive`. A transient sampling **error** maps to `unsupported_by_host` → `inconclusive`, never masking a genuine answer-vs-pattern mismatch (the regex path runs only on `status:'ok'`). Matches **P2 (harness, not agent)** and **P4 (facts before decisions)**. Request shape (`content:[{type:'image',data,mimeType},{type:'text'}]`, `maxTokens`) and response read (`content.text`) follow the MCP `sampling/createMessage` schema; 30s transport timeout, 512-token cap.

**Fixes folded in during review (2 independent code-review passes):**
- **P1 (ReDoS, raised on #1445):** the regex path becomes reachable here, so the raw `new RegExp(expected_pattern)` was swapped for `compileSafeRegex(...)` — defense-in-depth against catastrophic backtracking on the host's answer string. Semantics unchanged for safe patterns; rejections flow to the existing `invalid_expected_pattern` evidence.
- **P2 (this PR):** a non-text sampling content block previously yielded `answer = ''`, which a permissive pattern (`.`/`.*`) would *vacuously pass*. Now guarded to return `inconclusive` instead, with a regression test.

**Changes to reach merge-readiness.** Rebased onto `develop` to contain only the runtime-hook commit (the image_qa tool #1441 and DSL clause #1445 are already on `develop`); retargeted base from the feature branch to `develop`.

**Verification.** `tsc` build clean; 43 tests pass (`oc-assert-image-qa-runtime` 5 incl. the new regression case, `image-qa` contract 9, `image-qa` tool 6, `oc-assert` wiring suites). CI: ubuntu ×3, windows ×3, `agent-success-mock` green; macOS jobs were queue-delayed on GitHub's runners (platform-neutral change, no OS-specific code).

Merged to `develop`.

Issue #1457, opened 2026-05-29
Audit: SSOT (#1359) pillar gap analysis — primitives built, wiring/enforcement missing; develop↔main divergence


shaun0927:
## Summary (TL;DR)

A code-level audit of Pillars **A–D** against the product-direction SSOT (#1359) and the Portability-Harness Contract (`docs/roadmap/portability-harness-contract.md`) finds that **direction adherence is high, but achievement is gated by one recurring pattern**:

> **Primitives are built and unit-tested, but the wiring that connects them to the real call path — or the enforcement that makes them binding — is missing.** Many declared capabilities exist in code yet are *unreachable* (dead code) or *advisory-only* (cannot block).

On top of that, a **branch-divergence meta-issue** makes "how much is done" branch-dependent: the entire Pillar B (broker/lease/queue) stack lives on `main` but is **absent from `develop`**, the branch all new work builds on.

This issue catalogs the concrete gaps with evidence and proposes a phased action plan. Findings are an audit snapshot to be verified, not a regression report.

---

## Context

- Product-direction SSOT: #1359 (host-neutral MCP browser harness; 7 principles P1–P7; 6 pillars A–F)
- Normative contract: `docs/roadmap/portability-harness-contract.md`
- Recent merge energy is correctly aligned with the SSOT backlog (benchmark, skill memory, contracts, broker) — **drift is low**. The problem is *depth of completion*, not *wrong direction*.

---

## Meta-finding M0 — `develop` ↔ `main` divergence (highest priority)

The Pillar B stack landed on `main` but was never back-merged into `develop`:

```
src/session/target-lease-registry.ts   develop=absent  main=present
src/session/target-command-queue.ts    develop=absent  main=present
src/utils/controller-lock.ts            develop=absent  main=present
src/broker/discovery.ts                 develop=absent  main=present
src/transports/broker-proxy.ts          develop=absent  main=present
```

- `git rev-list --left-right --count origin/main...develop` → **31  106** (develop is 106 ahead, 31 behind; two-way divergence).
- `main` is simultaneously missing 106 commits of develop-side work (benchmark / skill / contract).
- Back-merge PR #1455 (`chore(release): back-merge main into develop`) is **OPEN / MERGEABLE** but unmerged.

[...]

PR #1460, opened 2026-05-29
feat(session): sliding idle-TTL target leases — reclaim crashed-client tabs (#1457 PR-3)

shaun0927:
Part of the SSOT (#1359) pillar gap audit (#1457) — **PR-3 / Pillar B backlog item 7**.

## The gap (lease expiry was dead code)
The lease registry already had `ttlMs` / `leaseExpiresAt` / `expire()`, and the auto-cleanup timer already called `expire()`. But both real `acquire()` call sites in `SessionManager.acquireTargetLease` passed **no `ttlMs`**, so `leaseExpiresAt` was never set → `expire()` was a permanent no-op. A disconnected/crashed client's leased tab was never reclaimed by lease TTL.

## Design: sliding idle TTL (not absolute)
A naïve absolute TTL would kill long-running agent tasks (SSOT open question #4). Instead the TTL **slides on activity**:

- `touch()` pushes `leaseExpiresAt = now + ttlMs` forward; `executeCDP` calls it on every command → an actively used tab is never reclaimed.
- A lease only reaches `expire()` when its owner has gone silent past the TTL — i.e. a disconnected/crashed client.

## Changes
- **`target-lease-registry.ts`** — store `ttlMs` on the record; `touch()` slides expiry; `acquire()` carries `ttlMs` across re-acquires.
- **`session-manager.ts`** — new `targetLeaseTtl` config (default 30 min; `0` disables). `acquireTargetLease` arms it but **exempts the `default` session** (mirrors the existing `sessionTTL` protection so single-agent workflows are never reclaimed). `executeCDP` touches the lease. The auto-cleanup `expire()` loop now **reclaims the orphaned tab** (best-effort `closeTarget`, skipping `preserve` leases) in addition to cancelling queued work.

## SSOT alignment
Pillar B / P3a. Default-off-equivalent for the common single-agent case (default session exempt); only multi-client/worker leases are subject to idle reclaim. No new deps.

## Verification
- `npx tsc --noEmit` → clean.
- `npm run lint:tier` → no core→pilot violation.
- `tests/target-lease-registry.test.ts` → 8 passed, incl. 3 new cases: active lease survives past the original deadline, a no-TTL (default-session) lease never expires, re-acquire carries+refreshes the TTL.

### DIFF
--- a/src/session-manager.ts
+++ b/src/session-manager.ts
@@ -53,6 +53,14 @@ export interface SessionManagerConfig {
   sessionTTL?: number;
   /** Auto-cleanup interval in milliseconds (default: 1 minute) */
   cleanupInterval?: number;
+  /**
+   * Idle TTL for a managed target lease, in ms (default: 30 minutes; 0 disables).
+   * Sliding — refreshed on every executeCDP call. A non-default-session lease that
+   * goes silent past this window is treated as a disconnected/crashed owner and its
+   * tab is reclaimed by auto-cleanup. The "default" session is exempt (mirrors the
+   * sessionTTL protection), and `preserve`-policy leases are never auto-closed.
+   */
+  targetLeaseTtl?: number;
   /** Enable auto-cleanup (default: true) */
   autoCleanup?: boolean;
   /** Maximum number of sessions (default: 100) */
@@ -99,6 +107,7 @@ export interface SessionManagerStats {
 const DEFAULT_CONFIG: Required<Omit<SessionManagerConfig, 'tenantManager' | 'strictTenantIsolation'>> = {
   sessionTTL: 30 * 60 * 1000,      // 30 minutes
   cleanupInterval: 60 * 1000,       // 1 minute
+  targetLeaseTtl: 30 * 60 * 1000,   // 30 minutes (sliding idle TTL; 0 disables)
   autoCleanup: true,
   maxSessions: 100,
   maxWorkersPerSession: 50,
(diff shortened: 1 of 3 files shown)


shaun0927:
## Review summary — MERGE-READY ✅ (after review follow-up commit)

Reviewed intent, SSOT alignment (#1359 Pillar B / P3a), and consistency with audit #1457 (**PR-3 / B7**), then ran an independent specialist code-review pass in an isolated worktree. The first pass found **two P1 issues**, both now fixed in `a564b3ec` and confirmed resolved by a second review pass.

**Intent & direction — valid.** The lease registry already had `ttlMs`/`leaseExpiresAt`/`expire()` and the cleanup timer already called `expire()`, but `acquire()` was never passed a `ttlMs`, so `leaseExpiresAt` was never set → `expire()` was a permanent no-op and a crashed client's leased tab was never reclaimed (audit B7). Arming a **sliding** idle TTL (refreshed on activity, not absolute) is the right design — it reclaims disconnected/crashed-owner tabs while protecting long-running agent tasks (SSOT open question #4). Default-session exemption + `0`-disables + `preserve`-policy skip keep the common single-agent path behavior-unchanged (P3).

**P1 findings fixed in follow-up:**
1. **(critical) Sliding signal was incomplete.** `touch()` was only called from `executeCDP`, but virtually every Puppeteer-based tool (navigate/interact/read_page/act/…) reaches the page via `getPage()`, which has **no** `executeCDP` call. An actively-used non-default-session tab would therefore still reach the idle sweep and be reclaimed mid-task — defeating the "sliding" intent and risking a P3 surprise-close. Fixed by touching the lease in `getPage()` too, immediately after the ownership check (guard-protected — a non-owner throws before the touch).
2. **`inherit()` TTL handling.** Used a truthy guard inconsistent with `acquire()`'s `!== undefined` style and never carried the parent's TTL absent an override, so a popup of a leased tab could become a never-expiring orphan lease. Fixed to carry parent TTL like the other inherited fields; default-session exemption preserved (a no-TTL parent still yields a no-TTL child).

**Verification (evidence-based):**
- `expire()` removes leases from the registry *before* returning them, so the subsequent best-effort `closeTarget()` → `release()` is a safe no-op (no double-free); the reclaim loop is correctly `await`ed and swallows errors.
- Map mutation during `expire()`'s `for…of` is well-defined; the reclaim loop iterates the returned array, not the live map.
- `tsc --noEmit` clean; `target-lease-registry` + `session-manager` suites green (35 passed, incl. 2 new `inherit` tests covering parent-TTL carry-through and explicit-override-wins).

**Findings:** P0: 0 · P1: 0 (2 fixed) · remaining P2s are non-blocking. Merge-ready once CI re-confirms on `a564b3ec`.

PR #1461, opened 2026-05-29
fix(cli): align `update --client` with supported setup clients; fail fast otherwise (#1457 PR-8)

shaun0927:
Part of the SSOT (#1359) pillar gap audit (#1457) — **PR-8 / Pillar A**.

## The dead path
`openchrome update --client cursor|windsurf|vscode` advertised those clients in `--client` help and forwarded to `openchrome setup --client X` — but `setup` only supports `claude|codex|opencode`, so the advertised value hit `process.exit(1)` mid-update. The flag also **omitted `opencode`**, which `setup` does support.

## Changes
- **cli/index.ts** — fix `--client` help to the real set: `(claude, codex, opencode)`.
- **cli/update-command.ts** — validate `--client` against `getSupportedMCPClients()` **before touching npm**; reject unsupported clients with a pointer to the README manual MCP config block (Cursor / VS Code / Windsurf) and the `--no-setup` escape hatch. Option type widened to a runtime-validated string.

## SSOT alignment
Pillar A (setup support for the declared host classes). Honest advertising: the CLI no longer offers automated setup it then rejects; manual config remains documented for the other clients.

## Verification
- `npx tsc --noEmit` → clean.
- `tests/cli/update-command.test.ts` → 6 passed, incl. `--client opencode` (regression guard) and fail-fast on `--client cursor` (npm not invoked, manual-config guidance printed).

### DIFF
--- a/cli/index.ts
+++ b/cli/index.ts
@@ -425,7 +425,7 @@ program
   .command('update')
   .description('Update the globally installed OpenChrome package')
   .option('--no-setup', 'skip MCP client reconfiguration after updating')
-  .option('--client <client>', 'MCP client to reconfigure after updating (claude, codex, vscode, cursor, windsurf)', 'claude')
+  .option('--client <client>', 'MCP client to reconfigure after updating (claude, codex, opencode)', 'claude')
   .option('--scope <scope>', 'Claude Code scope to use during setup (user or project)', 'user')
   .action((options) => {
     const code = runUpdateCommand({
--- a/cli/update-command.ts
+++ b/cli/update-command.ts
@@ -1,10 +1,16 @@
 import { spawnSync } from 'child_process';
+import { getSupportedMCPClients, isSupportedMCPClient } from './mcp-client-config';
 
 const UPDATE_COMMAND = ['npm', 'install', '-g', 'openchrome-mcp@latest'] as const;
 
 export interface UpdateCommandOptions {
   setup?: boolean;
-  client?: 'claude' | 'codex' | 'vscode' | 'cursor' | 'windsurf';
+  /**
+   * MCP client to reconfigure after updating. Validated at runtime against the
+   * automated-setup set (`getSupportedMCPClients()`); unsupported clients such as
+   * Cursor / VS Code / Windsurf are rejected with a pointer to the manual config.
+   */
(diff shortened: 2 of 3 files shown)


shaun0927:
## Review summary — MERGE-READY ✅

Reviewed intent, SSOT alignment (#1359 Pillar A), and consistency with audit #1457 (**PR-8 / A1**), then ran an independent specialist code-review pass in an isolated worktree.

**Intent & direction — valid.** `openchrome update --client cursor|windsurf|vscode` advertised clients that `setup` rejects (`claude|codex|opencode` only), so the advertised value hit `process.exit(1)` mid-update; the flag also omitted `opencode`, which `setup` *does* support. Fixing the help text + fail-fast validation against the single supported-client source is honest-advertising (Pillar A) and host-neutral — no host gets privileged automated setup it can't actually complete.

**Verification (evidence-based):**
- Guard `if (setup && options.client && !isSupportedMCPClient(options.client))` runs **before** npm is touched (no partial-update side effect); `--no-setup` correctly skips it; commander default `--client claude` passes unaffected.
- `getSupportedMCPClients()` / `isSupportedMCPClient()` derive from a single `SUPPORTED_CLIENTS = ['claude','codex','opencode']` constant in `cli/mcp-client-config.ts` — no second divergent list. `setup` genuinely handles `opencode` (writes `~/.config/opencode/opencode.json`), so the new path won't itself exit 1.
- `console.log` for progress / `console.error` for rejection is correct here — this is the standalone `update` CLI, not the MCP stdio server, so there's no JSON-RPC stdout to corrupt.
- README anchor `#quick-start` and the "Manual MCP config (Cursor / VS Code / Windsurf)" block both exist; the guidance is accurate.
- `tsc --noEmit` clean; `tests/cli/update-command.test.ts` 6 passed (incl. opencode regression guard + cursor fail-fast asserting npm is never invoked).

**Findings:** P0: 0 · P1: 0 · P2: 2 (non-blocking — a redundant `isSupportedMCPClient('opencode')` unit assertion, and the pre-existing `--client claude` omission in `getSetupArgs`; the behavior is already covered by `update-command.test.ts`). Merge-ready as-is.

PR #1468, opened 2026-06-02
docs(reliability): reconcile stale roadmap with shipped reality; record D4/D5

shaun0927:
## Summary

A post-ship reliability audit found the `docs/roadmap/` reliability docs were **stale**: all 7 implementation phases were marked unmerged though their PRs (#392/#395/#397/#398/#399/#400/#402) are all merged, and the "what is missing" motivation list still enumerated already-shipped capability. This PR reconciles the docs with shipped reality and promotes two implicit decisions to normative SSOT entries.

## Changes

- **`issue-reliability-tracking.md`** — flip 7 phase statuses to `[x] Merged (verified on main)`; reconcile the §7 E2E table to actual test files (`E2E-13..18` confirmed by explicit ID tag; `11/12/19/20/21` marked 🟡 filename-only, pending a tag — see PR-C follow-up); add a reconciliation note. Manual checklists §3–6 left **unchecked on purpose** — they are release-time process artifacts, not facts to assert.
- **`issue-reliability-guarantee.md`** — rewrite the "what is missing" list to past-tense **Delivered** with PR/file references; add a **Known Limitations** section (L1 retry-path consistency bug, L2 timeout cancellation trade-off).
- **`ssot-decisions.md`** — record **D4** (auto-retry = connection-error-only, at-least-once; at-most-once is the orchestrator's job) and **D5** (timeout/abort does not cancel the in-flight CDP call — orphaned background call accepted for never-hang). These previously lived only in code comments.

## Verification

- Documentation only; no behavior change. `main == develop` and the three docs were identical on both branches, so this applies cleanly.
- All 7 referenced PRs verified MERGED via `gh pr view`.

## Follow-ups (separate PRs)

- **PR-B** — fix L1: apply the timeout race + `reconcileAfterReconnect()` guard to the swallowed-error retry path in `src/mcp-server.ts` (per D4).
- **PR-C** — tag `E2E-11/12/19/20/21` in their test files and close the §7 table mapping.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1471, opened 2026-06-02
fix(reliability): make swallowed-error retry fire + guard it with reconcile/timeout (L1)

### LINKED ISSUE #1469: fix(reliability): swallowed-connection-error retry path lacks timeout race + reconcile guard (never-hang violation)
## Problem

Per SSOT decision **D4** (`docs/roadmap/ssot-decisions.md`), both auto-retry paths in `src/mcp-server.ts` must apply the same guards. They currently do not.

The **thrown-connection-error** retry path (~L2092) is careful:
- wraps the retried handler in a `Promise.race` timeout, and
- gates the retry on `sessionManager.reconcileAfterReconnect()` (aborts to the original error if reconciliation fails).

The **swallowed-connection-error** retry path (~L2148–2174) does **neither**:
- `~L2168`: `result = await Promise.resolve(tool.handler(...))` — **no timeout race**. In stdio mode (no outer HTTP request timeout) a hanging handler hangs the entire tool call → **violates the never-hang contract**. HTTP mode is incidentally masked by `HTTP_REQUEST_TIMEOUT_MS=30000`.
- **no `reconcileAfterReconnect()` gate** → exposed to the stale-target "wrong-target" errors the throw-path explicitly guards against.

## Fix

Mirror the throw-path guards in the swallowed-error path:
1. Wrap the retried handler in the same `Promise.race` timeout (`DEFAULT_TOOL_EXECUTION_TIMEOUT_MS`).
2. Gate the retry on `reconcileAfterReconnect()`; on reconcile failure keep the original error result.

No change to the happy path or to non-connection errors.

## Test

Add a stdio-mode regression: inject a swallowed connection error in a tool result, then a hanging handler on retry; assert the call returns (error) within the timeout rather than hanging.

## Refs

- SSOT decision **D4**, Known Limitations **L1** (`docs/roadmap/issue-reliability-guarantee.md`)
- Reconciliation PR #1468

shaun0927:
Fixes **L1** from the post-ship reliability audit. Closes #1469. Implements SSOT decisions **D4/D5** (recorded in #1468).

## The bug(s)

The swallowed-connection-error retry path in `MCPServer` — the branch that handles a tool which **returns** an `isError` result instead of throwing — had two defects:

1. **It was dead code.** The gate was `isConnectionError({ message: errorText })`. `isConnectionError()` runs its argument through `formatError()`, which only pulls `.message` off `Error` instances and otherwise does `String(value)`. A plain `{ message }` object stringifies to `"[object Object]"`, matching no connection pattern — so the retry **never fired**. Fixed by passing the `errorText` string directly.

2. **Missing guards vs. the thrown-error path.** Once activated, the retry must mirror the protections the thrown-error path (~L2092) already has:
   - a **`reconcileAfterReconnect()` gate** — abort the retry and keep the original error if reconciliation fails (stale target state would otherwise cause wrong-target errors);
   - a **`Promise.race` timeout** around the retried handler. A bare `await` could hang the **stdio** path indefinitely (no outer request timeout in stdio mode), violating the never-hang contract. (HTTP mode was incidentally masked by `HTTP_REQUEST_TIMEOUT_MS`.)

## Fix

`src/mcp-server.ts`: pass the string to `isConnectionError`, add the reconcile gate and the timeout race so both retry paths apply identical guards.

## Tests

`tests/mcp-server.test.ts` — 3 new regressions under **"swallowed connection error retry (L1)"**:
- retry fires and succeeds (path active, reconcile invoked);
- reconcile failure aborts the retry and keeps the original error;
- does-not-hang: retry bounded by the tool-execution timeout (fake timers).

`npm run build` clean; `tests/mcp-server.test.ts` green (35/35).

## Scope / safety

This activates a previously-dead retry path. Per **D4** it is connection-error-only, now reconcile-gated and timeout-bounded, so it is at-least-once and never-hang-safe. No happy-path or non-connection-error behavior changes.

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -2150,11 +2150,21 @@ export class MCPServer {
       // isError results, bypassing the thrown-error retry at the catch block above.
       if (result.isError && result.content?.[0]?.type === 'text') {
         const errorText = (result.content[0] as { text: string }).text;
-        if (isConnectionError({ message: errorText })) {
+        // Pass the string directly: isConnectionError() runs the value through
+        // formatError(), which only extracts `.message` from Error instances and
+        // otherwise calls String(value). A plain `{ message: errorText }` object
+        // therefore stringifies to "[object Object]" and matches no pattern — which
+        // had silently made this whole swallowed-error retry path dead code. (L1)
+        if (isConnectionError(errorText)) {
           console.error(`[MCPServer] Detected swallowed connection error in "${toolName}" result, attempting reconnect + retry`);
           try {
             const cdpClientRetry = getCDPClient();
             await cdpClientRetry.forceReconnect();
+            // Wait for session state reconciliation before retrying — mirrors the
+            // thrown-error retry path above. If reconciliation fails this throws and
+            // the catch below keeps the original error result, since stale target
+            // state would otherwise cause wrong-target errors. (SSOT decision D4 / L1)
+            await this.sessionManager.reconcileAfterReconnect();
             // Retry the tool call once
             const swallowedRetryContext: ToolContext = {
               startTime: Date.now(),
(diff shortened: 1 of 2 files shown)


shaun0927:
## Code review — merge-readiness analysis

**Direction / #1359 alignment:** ✅ This strengthens the *never-hang* guarantee of MCP tool calls — squarely within #1359's "host-neutral MCP browser harness … reliable tools" identity and the `issue-reliability-tracking.md` Responsibility Boundary ("per-tool-call result guarantee"). It serves both stdio and HTTP-daemon host classes. No new direction, no MCP-contract surface change.

**Correctness of the 3 original fixes:**
- Dead-code activation (`isConnectionError(errorText)` vs `{message}`) — verified: `formatError` only reads `.message` from `Error` instances, so the object form stringified to `"[object Object]"` and never matched. Correct.
- Reconcile gate before retry — mirrors the thrown-error path; on reconcile failure the outer catch keeps the original error result. Correct.
- Handler timeout race — closes the stdio never-hang gap. Correct.

**Behavior-change risk:** activating the previously-dead branch means a swallowed-connection-error result now triggers at most one reconnect + one retry (handler ≤2 calls, bounded, no loop). Consistent with the project's accepted **at-least-once** connection-retry semantics (D4); at-most-once remains the orchestrator's job.

**Review finding addressed:**
- **[P1] `forceReconnect()` was unguarded** in this path while the thrown-error path races it against `DEFAULT_RECONNECT_TIMEOUT_MS`. If Chrome were dead and the reconnect never resolved, the stdio path would hang *before* the retry — re-opening the window this PR claims to close. **Fixed in `25a98202`**: forceReconnect is now raced against the same timeout, and a 4th regression test ("does not hang when the reconnect itself stalls") proves it.
- **[P2] (reviewer)** "add `toHaveBeenCalledTimes(2)` to the handler-timeout test" — verified already present; no change.

**Tests:** 4 regressions under "swallowed connection error retry (L1)"; `tests/mcp-server.test.ts` 36/36 green; `npm run build` + eslint clean.

**Verdict:** merge-ready once CI re-passes on `25a98202`.

PR #1484, opened 2026-06-08
feat(broker): auto-elect lock winner as broker owner (S2 of #1480)

shaun0927:
## S2 of the #1480 PR program — owner auto-elect

**Stacked on #1479** (base `feat/1474-duplicate-controller-mcp-error`; review the
#1474 stack #1477 → #1478 → #1479 first). Policy gate: **S1 #1481** (D3 Q1′).

Wires the **owner half** of auto-elect coordinated sharing. Opt-in via
`--auto-elect` / `OPENCHROME_AUTO_ELECT=1`, **off by default**.

When enabled, the `serve --auto-launch` process that **wins** the controller lock
elects itself the broker **owner**:
- runs **dual transport** (`both`) so it serves its own host over stdio *and*
  exposes the broker over HTTP;
- publishes broker discovery metadata (`~/.openchrome/brokers/`) on a deterministic
  port `cdpPort + 200` (e.g. 9422 — clear of the `+100` headed-fallback) unless
  `--http`/`OPENCHROME_HTTP_PORT` is given.

S3 adds the **loser** half (attach as a client instead of failing fast); together
they restore stable parallel sessions through one coordinated owner — the
#1359/#1367 end-state.

### Design
- New pure module `src/broker/auto-elect.ts` holds the election decisions
  (`isAutoElectEnabled`, `shouldElectBrokerOwner`, `shouldClientAutoConnect`,
  `defaultBrokerHttpPort`) so they're unit-testable without a server/Chrome.
- `src/index.ts` consumes them in the existing transport/broker wiring. Explicit
  `--broker`/`--connect-broker` always take precedence over auto-elect.

### Safety
With `--auto-elect` unset, every touched branch is **byte-for-byte** the prior
fail-fast path — the default single-client experience is unchanged. This is the
opt-in mechanism; flipping it to the default is a separate, deliberate decision
once validated end-to-end (see D3 Q1′ Status).

### Verification
- `npm run build:src` ✓ · eslint ✓
- new `tests/auto-elect.test.ts` (14) ✓
- existing `broker-discovery` / `broker-lifecycle` / `controller-lock` suites (23) ✓
- `serve --help` lists `--auto-elect` ✓

> Note: the multi-process broker handshake (owner publishes → second session
> attaches) is exercised end-to-end by S3 + a smoke test; this PR is the owner
> publish path and its unit-tested decision logic. Part of #1480.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

## Goal

[...]

Issue #1480, opened 2026-06-08
Parallel multi-session is blocked by design — land the #1474 self-healing stack + make the opt-in broker path reachable (SSOT #1359 / audit #1457 alignment)


shaun0927:
> Investigation context: `openchrome-mcp` **1.12.7** (npm `latest`), Node v20.19.6, macOS (Darwin 25.3.0, arm64), MCP host = Claude Code with the global registration `openchrome serve --auto-launch` (stdio, default port `9222`, default profile `~/.openchrome/profile`). This issue consolidates a root-cause analysis with a **direction/SSOT alignment review** of the fixes already in flight, and catalogs additional wiring gaps found during the investigation. It is meant to sit under the SSOT (#1359) / audit (#1457) umbrella and to act as the tracking issue for the #1474 stack + follow-ups.

## TL;DR

- **Parallel multi-session is structurally blocked, by design — not a regression.** Since #1376 introduced the per-`(port, userDataDir)` controller-owner lock, N identically-registered sessions resolve to **1 working owner + (N−1) hard failures** surfaced to the host as a bare `Failed to reconnect to openchrome: -32000`.
- This is the **intended** safety posture: #1376 deliberately refuses a second *direct* controller because multiple direct controllers race on CDP target cleanup/reconnect (stale targets / MCP disconnects). The shared-owner replacement (`--broker`/`--connect-broker`, #1379) is **opt-in by recorded decision** (roadmap #1463 **D3**), so making the broker the default would contradict the SSOT.
- The genuine defects are therefore **(a) the default path deadlocks and gives an opaque error**, and **(b) the opt-in shared path is built but under-wired/under-discoverable**. The first is already addressed by the **#1474 stack (#1477 → #1478 → #1479, all OPEN, stacked, MERGEABLE)**. The second is the "primitives built, wiring/enforcement missing" pattern from the audit (#1457) and needs the follow-ups proposed below.
- **Ask:** land the #1474 stack on `develop`, then close the remaining wiring/discoverability gaps (G1–G4) so concurrent hosts have a *reachable, documented, portable* path — without changing the opt-in default that D3 froze.

---

## 1. Symptom & deterministic repro

With ≥2 concurrent host sessions sharing one global registration (`openchrome serve --auto-launch`, same port+profile):

- Every session except the first reports, repeatedly and deterministically: `Failed to reconnect to openchrome: -32000`.
[...]

PR #1494, opened 2026-06-09
feat(broker): auto-elect lock winner as broker owner

shaun0927:
Supersedes #1484 after its stacked base PR was squash-merged and the base branch was deleted.

## Goal
Implement #1480 S2: let the `--auto-launch` lock winner automatically become the local broker owner.

## Final Implementation Scope
- Add broker auto-election helper logic.
- Wire auto-election into the MCP server startup path.
- Add regression coverage for owner election and loopback unauthenticated HTTP for the auto-elected owner.

## Success Criteria
- A lock winner can start/own the broker without manual broker setup.
- Losing/duplicate controller behavior remains outside this PR's scope.
- The diff remains limited to S2 auto-election owner behavior.

## Verification Method
- Local targeted test: `npx jest tests/auto-elect.test.ts --runInBand`.
- Source build: `npm run build:src`.
- GitHub CI must pass before merge.

## Guardrails
- No client-side loser attach logic here; that is S3.
- No broker-owner death re-election here; that is S4.
- No changes to DuplicateController remediation semantics beyond consuming the already-merged base.

## Explicit Non-Goals
- Multi-controller sharing after the owner is already running.
- Re-election after broker loss.
- Broad auth model changes outside loopback behavior needed for the elected owner.

## Implementation Approach
Keep auto-election isolated in `src/broker/auto-elect.ts` and perform minimal wiring in `src/index.ts`.

## PR Decomposition
This is S2 of the #1480 stack and should merge before S3 and S4.

## Over-Engineering Checklist
- No new dependency.
- No generalized broker lifecycle manager.
- Only the owner-election path is added.

## Drift-Prevention Checklist
- Rebased onto current `develop` after #1493 was squash-merged.
- Net diff checked against `origin/develop`.
- S3/S4 descendants must be rebased after this PR lands.

## Language Boundary
No user-facing language expansion beyond necessary diagnostics/wiring.

## Critical Risk Review
Main risks are accidentally widening unauthenticated HTTP and changing non-auto-launch behavior. Tests cover auto-election and keep the feature scoped to the intended path.

## Definition of Done
- Local targeted tests pass.
- Source build passes.
- GitHub CI green.
- Squash-merged into `develop` before S3/S4 are retargeted.

### DIFF
--- a/src/broker/auto-elect.ts
+++ b/src/broker/auto-elect.ts
@@ -1,0 +1,87 @@
+/**
+ * Auto-elect coordinated sharing (#1480; SSOT decision D3 Q1′).
+ *
+ * Pure, side-effect-free decision helpers for the default `serve --auto-launch`
+ * election path:
+ *
+ *   - the process that WINS the controller lock becomes the broker OWNER and
+ *     publishes broker discovery metadata (so surplus sessions can attach);
+ *   - a process that LOSES to a healthy owner becomes a coordinated CLIENT
+ *     (`--connect-broker` proxy) instead of failing fast.
+ *
+ * The behavior is gated behind an explicit opt-in (`--auto-elect` /
+ * `OPENCHROME_AUTO_ELECT=1`). With the flag unset, the default path is byte-for-
+ * byte unchanged (fail-fast single owner), so this module's wiring has zero blast
+ * radius until an operator opts in. See docs/roadmap/ssot-decisions.md (D3 Q1′)
+ * for the policy and the eventual default-flip plan.
+ *
+ * Keeping the decisions here (rather than inline in src/index.ts) makes the
+ * election rules unit-testable without booting a server or Chrome.
+ */
+
+/** Offset from the CDP port used for the elected owner's broker HTTP endpoint. */
+export const BROKER_HTTP_PORT_OFFSET = 200;
+
+/**
+ * Is auto-elect enabled for this process?
+ *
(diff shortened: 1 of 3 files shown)


shaun0927:
Follow-up fix after CI failure:

## Root Cause
The S2 startup wiring passed `allowUnauthenticatedHttp: false` explicitly into the HTTP transport when `--allow-unauthenticated-http` was not set. That shadowed the transport-level `OPENCHROME_ALLOW_UNAUTHENTICATED_HTTP=1` fallback, so existing HTTP/both health-endpoint integration scenarios failed before the health endpoint could log its gating decision.

## Fix
Normalize the effective explicit opt-in in `src/index.ts` from both sources:
- CLI flag: `--allow-unauthenticated-http`
- Env: `OPENCHROME_ALLOW_UNAUTHENTICATED_HTTP=1|true|yes`

Then layer the S2 auto-elect exception on top only for the zero-config, loopback-only elected owner path.

## Guardrail Preserved
`HTTPTransport` remains the auth-policy SSOT. This patch only prevents CLI startup from passing a false value that suppresses the env opt-in.

## Verification
- `npm run build:src`
- `npx jest tests/integration/health-endpoint-gating.test.ts tests/auto-elect.test.ts --runInBand` → 21/21 passed
- `git diff --check`

Full GitHub CI rerun is pending on the updated head.

PR #1517, opened 2026-06-18
fix(gates): detect Cloudflare bot-check interstitials

changeroa:
## Summary
- Add a non-CAPTCHA bot-check gate signal for sparse CDN/WAF interstitials such as Cloudflare 'Just a moment...' pages.
- Surface bot_check through oc_gate_inspect instead of returning detected:false when no captcha iframe exists yet.
- Cover detector priority and oc_gate_inspect output with regression tests.

## Investigation
- Reproduced on Cloudflare Dashboard login: after reload the tab stayed on title 'Just a moment...' with only Cloudflare/Privacy links.
- Console capture returned zero page console entries, so the useful signal is DOM/title state rather than console errors.
- Existing blocking-page utility already knows 'just a moment', but oc_gate_inspect only composed captcha/SSO/paywall/2FA detectors and therefore returned detected:false for this gate.

## Test Plan
- npm test -- --runTestsByPath tests/gates/detect-other-gates.test.ts tests/tools/oc-gate-inspect.test.ts
- npm run build

### DIFF
--- a/src/gates/detect-other-gates.ts
+++ b/src/gates/detect-other-gates.ts
@@ -82,15 +82,18 @@ function botCheckProbe(): { selector: string } | null {
     return { selector: 'title:just-a-moment' };
   }
 
-  if (bodyText.includes('verify you are human') ||
-      bodyText.includes('checking if the site connection is secure') ||
-      bodyText.includes('review the security of your connection') ||
-      bodyText.includes('please stand by, while we are checking your browser') ||
-      bodyText.includes('bot protection') ||
-      bodyText.includes('automated access') ||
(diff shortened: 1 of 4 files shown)


shaun0927:
Reviewed and applied the small follow-up before merge: updated oc_gate_inspect docs/description for bot-check, tightened generic bot-check text so ordinary content is not enough by itself, and added positive/negative regression coverage.\n\nVerified locally:\n- npm test -- --runTestsByPath tests/gates/detect-other-gates.test.ts tests/tools/oc-gate-inspect.test.ts\n- npm run build

